PR #42, opened 2021-01-27
Fix large files in vscode extension

lweehuizen-rbx:
Wait for child process to exit before resolving formatting promise so that large files are fully processed. Previously large files were being cut off as the data callback can be called multiple times and the current implementation was only allowing one callback.

Wrap path in quotes, so that paths with spaces still work when invoking exec. On Mac the default location of the StyLua binary will be under `Application Support` when installed via VSCode. This causes exec to fail due to the space. Wrapping in quotes solves the issue.

### DIFF
--- a/stylua-vscode/CHANGELOG.md
+++ b/stylua-vscode/CHANGELOG.md
@@ -11,6 +11,10 @@ To view the changelog of the StyLua binary, see [here](https://github.com/Johnny
 
 ## [Unreleased]
 
+### Fixes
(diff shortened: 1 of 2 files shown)


JohnnyMorganz:
Thanks for the PR! It looks good to me at first glance, but I'll test it out later. Could you add the fixes to the extension changelog?

lweehuizen-rbx:
Thanks @JohnnyMorganz. I've updated the Changelog.

JohnnyMorganz (approved):
Thanks!

PR #84, opened 2021-03-21
Long assignments are put on newline if not hangable

### LINKED ISSUE #62: Long assignments should be put on new line if they can't hang
Something like
```lua
local LoadAddOn, UnitName, GetRealmName, UnitRace, UnitFactionGroup, IsInRaid = LoadAddOn, UnitName, GetRealmName, UnitRace, UnitFactionGroup, IsInRaid
```
... should be formatted as ...
```lua
local LoadAddOn, UnitName, GetRealmName, UnitRace, UnitFactionGroup, IsInRaid =
    LoadAddOn, UnitName, GetRealmName, UnitRace, UnitFactionGroup, IsInRaid
```

JohnnyMorganz:
Closes #62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 For the auto styles, we will prefer the quote type specified, but fall back to the opposite if it means there are fewer escapes. For the
 force styles, we will always use the quote type specified.
 - StyLua will now error when unknown fields are found in the configuration `stylua.toml` file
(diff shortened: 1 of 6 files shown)


JohnnyMorganz:
Currently doesn't indent correctly, if we are already in a higher indent level.
Also, what should we do in cases where we have a single, long argument, but it goes over the column_width by a small margin (see: the current failed test)?

Test fails on
```lua
local MODULE_NOT_READY_MESSAGE = "API has not been fetched yet; try using API.isReady() before calling API functions"
```
StyLua wants to change it to:
```lua
local MODULE_NOT_READY_MESSAGE =
    "API has not been fetched yet; try using API.isReady() before calling API functions"
```
but does this make much of a difference?

PR #156, opened 2021-05-14
Function call expanding

### LINKED ISSUE #148: Long function chain where only single call in function args goes over limit
```diff
-                       jestExpect(
-                               ReactIs.typeOf(React.createElement(React.Profiler, {id="foo", onRender=jest:fn()}))
-                       ).toBe(ReactIs.Profiler)
-                       jestExpect(
-                               ReactIs.isProfiler(React.createElement(React.Profiler, {id="foo", onRender=jest:fn()}))
-                       ).toBe(true)
+                       jestExpect(ReactIs.typeOf(React.createElement(React.Profiler, { id = "foo", onRender = jest:fn() }))).toBe(ReactIs.Profiler)
+                       jestExpect(ReactIs.isProfiler(React.createElement(React.Profiler, { id = "foo", onRender = jest:fn() }))).toBe(true)
```
Goes to 146 characters, over column limit
Originally from #144 @matthargett

JohnnyMorganz:
Fixes #148 - allows expanding of calls which contain single function args.

This PR also uses an infinite width shape for the first iteration of formatting arguments, so that they line up on a single line as much as possible - making it easier to determine whether we need to wrap.
It also uses `.take_last_line()` instead of `.to_string().len()`, as some values are function calls with a multiline arg inside of them.
Also checks the shape after the loop through arguments is complete, as we may be over budget.

Still need to decide on one thing: might want to make the function args formatting as a whole simpler and more akin to prettier - where we expand function args onto multiple lines if there is a multiline arg present within it (excluding if the multiline arg is the last arg/only arg).
We will then, however, need to add special cases for things like `Roact.createElement()` or `setmetatable`, as they are normally inlined rather than expanded. Going to think about this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Long prefix expressions which are hangable and go over the line limit (e.g. `("foooo" .. "barrrrrrr" .. "bazzzzzz"):format(...)`) will now hang multiline ([#139](https://github.com/JohnnyMorganz/StyLua/issues/139))
-- Function calls with single arguments are now possible to be expanded. This will allow the call to be expanded if the line goes over budget.
+- Function calls with single arguments are now possible to be expanded. This will allow the call to be expanded if the line goes over budget. ([[#156](https://github.com/JohnnyMorganz/StyLua/issues/156)])
 
 ### Fixed
 - Fixed 1 or 2 digit numerical escapes being incorrectly removed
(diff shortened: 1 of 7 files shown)


JohnnyMorganz:
The failing test is because of the change in using `.take_last_line()` rather than `.to_string().len()`. The last line is technically only a few characters now, but the argument was multiline - this is the decision that needs to be made (and in doing so, will unify the formatting more to the style of prettier)

matthargett (approved):
Yay for the new test!

JohnnyMorganz:
A common change in formatting occuring from this update (as explained earlier), is that e.g.
```lua
table.insert(
    Messages,
    Roact.createElement(RadioMessage, {
	    LayoutOrder = Index,
	    Message = Message,
    })
)
```
is getting formatted as
```lua
table.insert(Messages, Roact.createElement(RadioMessage, {
    LayoutOrder = Index,
    Message = Message,
}))
```
I personally do not mind this, but I'm unsure whether it will make more complex items harder to read - it is still under the column width though, and if it went over then it would revert to the original.
Again, this could be fixed with the change described in OP, but not sure if its something we want to commit to

JohnnyMorganz:
Going to accept this change for the time being - may look into the alternative at a later time

Issue #158, opened 2021-05-15
Incorrect indentation of hanging prefix expressions


JohnnyMorganz:
Bug from #139 - this is currently unreleased
```lua
		local CreatedDirection = (
			DirectionalCF * CFrame.fromOrientation(0, 0, RNG:NextNumber(0, TAU)) * CFrame.fromOrientation(
				math.rad(RNG:NextNumber(
					GunMainConfiguration.BulletMinSpreadAngle or GlobalConfiguration.DEFAULT_MIN_SPREAD_ANGLE,
					GunMainConfiguration.BulletMaxSpreadAngle or GlobalConfiguration.DEFAULT_MAX_SPREAD_ANGLE
				)),
				0,
				0
			)
		).LookVector
```
... gets formatted as ...
```lua
		local CreatedDirection = (
				DirectionalCF
				* CFrame.fromOrientation(0, 0, RNG:NextNumber(0, TAU))
				* CFrame.fromOrientation(
				math.rad(RNG:NextNumber(
					GunMainConfiguration.BulletMinSpreadAngle or GlobalConfiguration.DEFAULT_MIN_SPREAD_ANGLE,
					GunMainConfiguration.BulletMaxSpreadAngle or GlobalConfiguration.DEFAULT_MAX_SPREAD_ANGLE
				)),
				0,
				0
			)
			).LookVector
```
i.e. note the lack of indentation in the prefix expression

JohnnyMorganz:
This was actually fully fixed from #160 - added the test in https://github.com/JohnnyMorganz/StyLua/commit/8caacbd476c40deba28364aeab8d8a71da0a1590

PR #160, opened 2021-05-16
Move indentation level to Shape

### LINKED ISSUE #153: Incorrect mutable state persisting between formatting tactics
Internally in StyLua, the way some formatting works is by following a series of choices, and seeing which fits best. For example, for Assignment/LocalAssignment, we firstly attempt to format onto a single line (setting column width to infinite). If this doesn't fit, we then see if its hangable and format that. If this fails, we just format it normally, and hang at the equals token if possible.

In the end, this is 3 different formatting tactics tested internally. However, there is some mutable state incorrectly persisting between all of them: the `indent_ranges` HashSet in `Context` (through `ctx.add_indent_range` and `ctx.get_range_indent_increase`). If the first formatting tactic caused a range to be adding to the extra `indent_ranges` HashSet, this persists onto any subsequent formatting tactics, even if the original formatting tactic was discarded. In most cases, this is normally fine (as subsequent formatting tactics will normally also add this same range again, and since its a HashSet, only one copy of the range persists), but this caused the issue in #151. In #151, on the first tactic run-through, we retrieve any current indent increase for the function call and increment it by one, then we add the function call range into the HashSet. On the second tactic run-through, we retrieve the range indent increase again - but this is one too many, because now the function call range is also included, which wasn't present the first time, and then we increment it by one again.
#152 provides a fix for this issue through two ways:
1) stop a second tactic running because the first tactic was actually OK in this case
2) adding the function call range to the HashSet before calculating the range indent increase, and removing the +1 increment outside of it.

This, however, jumps around the main issue.

Ultimately, everything does seem to be formatting OK for now, but this issue can cause unwanted side effects in the future. This should probably be patched, e.g. by any of the following options:
1) Making a copy of the context before attempting any tactic, and provide this same, original context to each one. No state will persist between them
2) Stop providing a mutable reference to Context in each formatter - create an immutable Context for each one [this is heavy though]
3) Move `indent_ranges` to `Shape`, or add something similar to it (see `Indent` in rustfmt `Shape`). Shape is already an immutable object being used for formatting, so it makes sense here. However, `Shape` is currently lightweight since it is a bunch of primitives, so it can implement `Copy`. Migrating `indent_ranges` as-is will add a HashSet, which cannot be Copied, only Cloned, which is performance hit.

JohnnyMorganz:
This PR creates a new struct `Indent` which is part of `Shape` (inspired by rustfmt). The `Indent` struct contains information about both the block indent level, and any additional indent level (additional caused by e.g. multiline table bodies). Like the `Shape` struct, the `Indent` struct is based around immutable data structures

This PR migrates away from "indent ranges" stored in context, and fixes #153. This should hopefully also simplify logic for future development, as there is no mysterious indent range which led to weird bugs - the exact indent level is now stored inside the Shape/Indent struct, and is constant. Context is also now no longer mutable.


JohnnyMorganz:
Weird and annoying bug, caused when we have an expression being hung, at a specific hang level.
E.g.
```lua
fooooo
    + barrrrr
    + bazzzz
```
or in a wider context
```lua
local x = fooooo
    + barrrrr
    + bazzz
```
In particular, the leftmost expression of the binary operation (`fooooo`) has a different indentation level to the rest, as it is being inlined on the same line, rather than on a new line. However, with this transition to indentation via shape, the additional indent level covers the whole binary expression (`fooooo + barrrrr + bazzz`). Need to find a way to reduce the additional indent level of the leftmost expression by one in these situations.

In particular, need to notice that there are other situations where this does not occur (mainly in conditions):
```lua
if
    fooooo
    and barrrrr
    and bazzz
then
    ...
end
```
There is no hang level here.

Need to find a way to fix this, it is proving quite challenging currently.

JohnnyMorganz:
This actually also partially fixes #158 :)
Hopefully this should make it much easier to reason about what the given indent level is at an exact moment in time, without relying on ranges (which also got purged when making new AST nodes)

PR #203, opened 2021-06-20
Add `--verify` to detect code correctness changes

### LINKED ISSUE #199: Add a `--verify` flag to check the output AST still matches the input AST
Would help catch mistransformations in code, especially ones where the syntax is still valid.
StyLua shouldn't be modifying the output AST (stripped of its whitespace), so it can help flag things up. full-moon has a `.similar()` method which we could use here.

JohnnyMorganz:
When this flag is enabled, StyLua will completely reparse the generated output and checks it against the input provided. If any changes are detected, we error for formatting that file and skip it.
Closes #199


JohnnyMorganz:
Added an AST massager in order to normalise input and output AST better so that we don't trigger false positives when checking for similarity due to expected changes (e.g. string quote changes etc.)
Few things left to go:
- Can we reduce the amount of "normalisation" happening in the massager? Especially for expressions: we just remove all parentheses completely, which may miss errors we wouldn't otherwise catch.
- There are two false positives still triggering but I can't quite work out *what* they are yet :) need to find those and fix them

`--verify` should be treated as an aide rather than being completely reliant on it. It gives users (and me!) some peace of mind when migrating a large project, but there are things it still may not catch. It should act as a guide to indicate that you may want to pay special attention to anything it flags, but human eyes are the best at catching these things.

PR #207, opened 2021-06-21
Add extension version specification

### LINKED ISSUE #196: Add option to specify StyLua version from extension
This would be extremely useful in the case of a bug in the current version. Currently, I have to resort to manually finding the default StyLua path and inserting the old version from Github. Or temporarily changing the StyLua directory to an older version that I've downloaded.
Regardless, this is a useful feature.

LastTalon:
Extension users will be able to select the release version used by the extension.

Versions options include all minor versions after v0.5 as well as a `latest` option. This allows an easy choice of version with the additional option of always updating to the latest version regardless of changes. A target version selector is included for users who need specific versions installed.

This resolves #196.

### DIFF
--- a/stylua-vscode/package.json
+++ b/stylua-vscode/package.json
@@ -42,19 +42,6 @@
     "configuration": {
       "title": "StyLua",
       "properties": {
-        "stylua.management": {
-          "type": "string",
-          "default": "partial",
-          "enum": [
-            "partial",
-            "none"
-          ],
-          "enumDescriptions": [
-            "Manage StyLua installation just for this extension.",
-            "Do not manage StyLua installation."
-          ],
-          "description": "The management style of the StyLua installation."
-        },
         "stylua.releaseVersion": {
           "type": "string",
           "default": "latest",
@@ -64,18 +51,9 @@
             "v0.8",
             "v0.7",
             "v0.6",
-            "v0.5",
-            "v0.4",
-            "v0.3",
-            "v0.2",
-            "v0.1"
+            "v0.5"
           ],
           "description": "The release version to download."
-        },
-        "stylua.styluaBinary": {
-          "type": "string",
-          "default": "stylua",
-          "description": "The name or path of the StyLua binary."
         },
         "stylua.styluaPath": {
           "type": [
--- a/stylua-vscode/package.json
+++ b/stylua-vscode/package.json
@@ -56,7 +56,7 @@
           "enumDescriptions": [
             "The most recent version released. This will always keep you up to date."
           ],
-          "description": "The release version to install."
+          "markdownDescription": "The release version to install. This is overridden by `#stylua.targetReleaseVersion#`."
         },
         "stylua.targetReleaseVersion": {
           "type": "string",
--- a/stylua-vscode/package.json
+++ b/stylua-vscode/package.json
@@ -53,7 +53,15 @@
             "v0.6",
             "v0.5"
           ],
-          "description": "The release version to download."
+          "enumDescriptions": [
+            "The most recent version released. This will always keep you up to date."
+          ],
+          "description": "The release version to install."
+        },
+        "stylua.targetReleaseVersion": {
+          "type": "string",
+          "default": "",
+          "markdownDescription": "Target a specific release version tag such as `v0.9.1` or `v0.9`. This overrides the version set by `#stylua.releaseVersion#`."
         },
         "stylua.styluaPath": {
           "type": [
--- a/stylua-vscode/package.json
+++ b/stylua-vscode/package.json
@@ -42,6 +42,41 @@
     "configuration": {
       "title": "StyLua",
       "properties": {
+        "stylua.management": {
+          "type": "string",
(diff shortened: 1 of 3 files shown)


LastTalon:
Amended with correct author information.

JohnnyMorganz:
I was thinking, is the `management` configuration necessary?
The `styluaBinary` configuration is optional. Could it not just work so that if that is empty, then the extension manages the binary. If it is filled out by the user, then don't manage?

stylua-vscode/src/util.ts
```diff
@@ -21,8 +22,20 @@ type GithubRelease = {
   html_url: string;
 };
 
-const getLatestRelease = async (): Promise<GithubRelease> => {
-  return await fetch(RELEASES_URL).then((r) => r.json());
+const getRelease = async (version: string): Promise<GithubRelease> => {
+  if (version === "latest") {
+    return await (await fetch(RELEASES_URL + "/latest")).json();
```
LastTalon, line 27: This is the only place that we modify the URL. We could update the worker to handle this. Although, I would still like to add github authentication in a separate PR.

stylua-vscode/package.json
```diff
@@ -42,6 +42,19 @@
...
+        "stylua.releaseVersion": {
+          "type": "string",
+          "default": "latest",
+          "enum": [
+            "latest",
+            "v0.9",
+            "v0.8",
+            "v0.7",
+            "v0.6",
+            "v0.5"
```
LastTalon, line 54: We could include older versions in the future by adding support for stylua versions that don't support ranges.

LastTalon:
The functionality here should be finished. I would just like to add a configuration option for selecting a specific release tag for cases where a specific version is needed, such as working on a project that requires it, or if a new version introduces a bug.

stylua-vscode/package.json
JohnnyMorganz: Would it be better for this to be a text box rather than a drop down list? There are lots of versions, and more will be added in the future, so the pro here is that it won't get really long, and we don't need to keep remembering to update the extension as more versions are released.
The con is that it means the user needs to input the correct version specifically, which they will need to remember/find from GitHub.

LastTalon: This is what I was referencing in my final comment that I'll be adding today. I figured for the average user the dropdown is easier to select a correct version with, but a string to select a specific release tag as an override is good for users with a specific version need including regressions, breaking bugs, or pre-releases.

This also lets us control when older version are no longer being supported officially by the extension (such as versions prior to v0.5 currently). Overall I think this is a good tradeoff. Let me know your thoughts.

stylua-vscode/package.json
```diff
@@ -42,6 +42,27 @@
...
+          ],
+          "enumDescriptions": [
+            "The most recent version released. This will always keep you up to date."
```
JohnnyMorganz, line 57: I know its already mentioned in `stylua.targetReleaseVersion`, but is it also worth saying something along the lines of "This setting is overriden by #stylua.targetReleaseVersion# if provided", for completeness sake?

stylua-vscode/src/util.ts
```diff
@@ -143,10 +166,11 @@ export const ensureStyluaExists = async (
...
 
     try {
-      const version = (await executeStylua(path, ["--version"]))?.trim();
-      const release = await getLatestRelease();
+      const currentVersion = (await executeStylua(path, ["--version"]))?.trim();
+      const desiredVersion = getDesiredVersion();
+      const release = await getRelease(desiredVersion);
       if (
-        version !==
+        currentVersion !==
```
JohnnyMorganz, line 173: I think we still need a way to let the user know that there is a new *latest* release to StyLua, even if they aren't set to latest. It can keep them up to date.
But this will keep showing every time the extension becomes activated, so there should be an option along the lines of "dont remind me for this version", where we save a config to not show the prompt again if the user ignored it.

JohnnyMorganz (commented):
Two small things.
I think it is important we still let the user know that a new release has happened for StyLua, to keep them up to date, even if releaseVersion is not set to latest.

stylua-vscode/package.json
LastTalon: I looked at other extensions and chose this way because others don't seem to reflect settings this way. It makes sense that a user normally using `releaseVersion` would not generally care about `targetReleaseVersion` while setting it, but the reverse is always true. That's just my rationale, I'll leave this final decision up to you. Let me know and I'll make the change.

JohnnyMorganz: Its an incredibly minor point, so doesn't really matter if we don't add it, but I think it helps in the unlikely event that a user forgets that they have a targetReleaseVersion set - it just points them to the setting in case they are wondering why it isn't taking affect.

stylua-vscode/src/util.ts
LastTalon: This is a good concept, but quickly became out of scope with several pitfalls. I've created #216 to track this.

PR #208, opened 2021-06-21
Fix severe performance regression due to change of table formatting

### LINKED ISSUE #205: Changes to table constructor formatting severely degraded performance
In https://github.com/JohnnyMorganz/StyLua/commit/fbe55ebf67141ecdda4a6fb4d34b6e118486a044 (https://github.com/JohnnyMorganz/StyLua/pull/182), we changed the formatting tactic for table constructors.
Originally, to see if a table constructor is over the column width, we would check the difference between the position of the start and end brace, and add it to the current width taken. This was quite naiive, because if you had a badly formatted table constructor with lots of spaces, it would go over the width.

To change this, we first attempt to format the whole table on a single line. If it then goes over width, then we stop, scrap the formatting we did, and then format on multiple lines. For code with lots of (nested) tables, this degraded performance significantly, as we double the amount of (quite expensive) work done per table, which adds up exponentially as you increase the depth of a table.

Reverting to a snapshot before this commit heavily improves performance, therefore a new heuristic needs to be defined for this.
We could either revert to the tactic we had originally (which still isnt great), or we could find something else.
Maybe, rather than attempting to format the whole table singleline then check if its over width, we format a small chunk at a time, examine how far we are, and bail early. This short circuiting should reduce the amount of operations done significantly.

JohnnyMorganz:
Fixes #205, which gives an explanation of what happened.

Our change in 0.9.0, as highlighted in #205, led to an exponential time complexity in formatting deeply nested tables. This is because we attempt to format the table on a single line to see if it is over width. This means, if the table contains more tables, we attempt this *again* for each table field, and this adds up exponentially the deeper the table.

Two solutions were attempted here:
1)  Revert back to our original, where we check the position between the opening and closing brace in the input. This is the cheapest solution (taking O(1) time), but isn't great since we rely on the input: if the input is badly formatted, it will lead to the table being pushed multiline when it could've been collapsed.
2) Implement a fail fast version of what we attempted in 0.9.0. Instead of formatting the whole table on a single line, then checking its width to determine if its over budget, we format a single field at a time and see if it has gone over the column width - if it has, bail early and expand multiline.

This PR adds a `nested_tables.lua` benchmark. Benchmarking this on current trunk took **3.0536s** for this single file.
Solution 1) leads to a performance improvement of **98.5%**, going from 3.0536s -> **44.26ms**
Solution 2) leads to a performance improvement of **40.0118%**, going from 3.0536s -> **1.8318s**. This however, is a performance **regression** of **4038.1%** compared to solution 1 (i.e. v0.8.2). Solution 2 is clearly still exponential (which makes sense, if we have tables early on in the fields).

We will revert back to solution 1, as present in v0.8.2. This is the best solution by far, and it was never actually flagged as problematic by any users before 0.9.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fixed severe performance regression due to a change in table formatting leading to exponential blowup for nested tables. ([#205](https://github.com/JohnnyMorganz/StyLua/issues/205))
 
 ## [0.9.2] - 2021-06-20
 ### Changed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,3 +45,7 @@ insta = { version="1.7.1", features=["glob"] }
 [[bench]]
 name = "date"
 harness = false
+
+[[bench]]
+name = "nested_tables"
+harness = false
--- a/benches/nested_tables.lua
+++ b/benches/nested_tables.lua
@@ -1,0 +1,464 @@
+return function()
+	foo(function()
+		local foo = {
+			foooooooBar = true,
+			barrrrrrrr = false,
+			bazzzzzzzzzzzz = false
+		}
+
+		local data = {
+			foooooooobarrr = {
+				bazzzzzzzzing = {
+					abadndafffffffff = "baszzzzzzzzzzfooooo",
+				},
+				singalong = nil,
+				configurationn = nil,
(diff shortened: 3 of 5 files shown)


JohnnyMorganz:
Here are some example flamegraphs to show the performance impact of what we had:
Current trunk (v0.9.2):
flamegraph-0 9 2-nested_tables
Using brace ranges (Solution 1):
flamegraph-braces_range-nested_tables
Using fail fast (Solution 2):
flamegraph-fail_fast-nested_tables

As can be seen in 0.9.2 and solution 2, the majority of the time is taken up by `format_table_constructor`, and `format_expression_internal -> format_table_constructor`. This is due to the exponential blowup for formatting fields inside a table, which are themselves table expressions.

PR #232, opened 2021-07-24
Collapse guard statements on a single line

### LINKED ISSUE #161: Format single conditional + LastStmt (`return/break`) "guard" statements on a single line?
Its common to use single conditional statements + returns to act as guards/short circuits in code:
```lua
function foo(a, b, c)
    if a == "foo" then return true end
    if a == "bar" then return false end
    if not b then return end

    -- ...

    for i,v in pairs() do
        if calculateSomething(i) then break end
        -- ...
    end
end
```

Currently, however, StyLua will expand each conditional block with something like:
```lua
function foo(a, b, c)
    if a == "foo" then
        return true
    end
    if a == "bar" then
        return false
    end
    if not b then
        return
    end

    -- ...
end
```
which might be making it unnecessarily longer and more verbose. Having them on a single line seems cleaner and more elegant.
Should we keep these "guard" conditions on a single line? Prettier also allows this with `if (foobar) return;` etc.

Some points that we will enforce if this happens:
- We will only allow inlined `if condition then ... end` blocks. If the if statement contains an `elseif` or `else`, it will not be inlined.
- No other blocks (`while`/`do`/`for ... do ... end`) will be formatted as such, it does not make sense to do so.
- The body of the if statement must only be a LastStmt (`return/break/continue`) - `if condition then x = x + 1 return end` is not allowed.
- If the whole statement (`if condition then return end`) surpasses the column width, we will not inline the statement.
- If the `condition` in `if condition then` spans multiple lines, we will not inline the statement.
- If the return value(s) in `if condition then return ... end` span multiple lines, we will not inline the statement.
- If the statement contains any internal comments (not leading/trailing the whole statement), we will not inline the statement.

JohnnyMorganz:
**WIP: still need to determine whether this PR is reasonable or not to use**

This PR formats guard-style if statements on a single line:
```lua
function foo(a, b, c)
    if a == "foo" then return true end
    if a == "bar" then return false end
    if not b then return end

    -- ...

    for i,v in pairs() do
        if calculateSomething(i) then break end
        -- ...
    end
end
```
#161 provides more information on the change.

Closes #161

### DIFF
--- a/src/formatters/block.rs
+++ b/src/formatters/block.rs
@@ -30,22 +30,13 @@ macro_rules! update_first_token {
 }
 
 pub fn format_return(ctx: &Context, return_node: &Return, shape: Shape) -> Return {
-    // Calculate trivia
-    let leading_trivia = vec![create_indent_trivia(ctx, shape)];
-    let trailing_trivia = vec![create_newline_trivia(ctx)];
-
     if return_node.returns().is_empty() {
-        let token = fmt_symbol!(ctx, return_node.token(), "return", shape).update_trivia(
-            FormatTriviaType::Append(leading_trivia),
-            FormatTriviaType::Append(trailing_trivia),
-        );
+        let token = fmt_symbol!(ctx, return_node.token(), "return", shape);
 
         Return::new().with_token(token)
     } else {
-        let token = fmt_symbol!(ctx, return_node.token(), "return ", shape)
-            .update_leading_trivia(FormatTriviaType::Append(leading_trivia));
-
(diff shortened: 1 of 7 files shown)


matthargett:
cc @magimaster @zotethemighty

MagiMaster:
> cc @MagiMaster @ZoteTheMighty

I put my thoughts on the issue itself, so the comments here can be about the implementation details.

JohnnyMorganz:
Closing as this branch is quite stale now.
May look into it in future, behind an option flag. Will not be a default though

PR #241, opened 2021-08-16
Extension GitHub authentication

### LINKED ISSUE #218: extension: Add ability to provide GitHub auth token
#207 changes from using the cloudflare worker to interfacing with the GitHub API directly to get releases.
However, using the API as a guest is subject to very stringent rate limits. There should be the ability to provide an auth token (via a command + a show a button if there is ever a request error)

LastTalon:
Extension users will be able to authenticate with github through vscode and will be prompted to when hitting the rate limit. No special scopes/permissions are needed as the extension only accesses public data.

This resolves #218 and allows proper usage of changes in #207 by allowing users to avoid REST API limits.

This change isn't that complicated, as vscode provides a nice interface for authentication providers and provides a built in github authentication provider. However, doing this nicely, and for future maintenance sanity, does require a fair bit of code reorganization.

- [x] Separate github interface into its own file
- [x] Separate downloading interface into its own file
- [x] Add authentication functionality
- [x] Prompt for authentication

### DIFF
--- a/stylua-vscode/src/download.ts
+++ b/stylua-vscode/src/download.ts
@@ -1,0 +1,142 @@
+import * as vscode from "vscode";
+import * as unzip from "unzipper";
+import * as util from "./util";
+import fetch from "node-fetch";
+import { createWriteStream } from "fs";
+import { executeStylua } from "./stylua";
+import { GitHub, GitHubRelease } from "./github";
+
+export class StyluaDownloader {
+  constructor(
+    private readonly storageDirectory: vscode.Uri,
+    private readonly github: GitHub
+  ) {}
+
+  public async ensureStyluaExists(): Promise<string | undefined> {
+    const path = await this.getStyluaPath();
+
+    if (path === undefined) {
+      await vscode.workspace.fs.createDirectory(this.storageDirectory);
+      await this.downloadStyLuaVisual();
+      return await this.getStyluaPath();
+    } else {
+      if (!(await util.fileExists(path))) {
+        vscode.window.showErrorMessage(
+          `The path given for StyLua (${path}) does not exist`
+        );
+        return;
+      }
+
+      try {
+        const currentVersion = (
+          await executeStylua(path, ["--version"])
+        )?.trim();
+        const desiredVersion = util.getDesiredVersion();
+        const release = await this.github.getRelease(desiredVersion);
+        if (
+          currentVersion !==
+          `stylua ${
+            release.tagName.startsWith("v")
+              ? release.tagName.substr(1)
+              : release.tagName
+          }`
+        ) {
+          this.openUpdatePrompt(release);
+        }
+      } catch (err) {
+        vscode.window.showWarningMessage(
+          `Error checking the selected StyLua version, falling back to the currently installed version:\n${err}`
+        );
+      }
+
+      return path;
+    }
+  }
+
+  public downloadStyLuaVisual(): Thenable<void> {
+    return vscode.window.withProgress(
+      {
+        cancellable: false,
+        location: vscode.ProgressLocation.Notification,
+        title: "Downloading StyLua",
+      },
+      () => this.downloadStylua()
+    );
+  }
+
+  private async downloadStylua(): Promise<void> {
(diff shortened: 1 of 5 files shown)


LastTalon:
The changes are currently non-functional because of the reorganization, so they can't currently be tested. I'll get back when changes are testable and I've done my own testing.

However, keeping an eye at the moment and doing light review of the changes could speed up the final review because things have moved around.

LastTalon:
The parts I tested seem to be working fine at this point. It just needs a prompt to occur when the actual rate limiting happens while not authenticated.

stylua-vscode/src/download.ts
```diff
@@ -0,0 +1,154 @@
...
+            "Authenticating with GitHub can fix rate limits.",
+            "Authenticate with GitHub"
+          );
+          switch (option) {
+            case "Authenticate with GitHub":
+              if (await this.github.authenticate()) {
+                return this.ensureStyluaExists();
+              }
+          }
+        }
```
JohnnyMorganz, line 61: Should we just have a single prompt here, rather than two? It can get quite cluttered imo.

stylua-vscode/src/download.ts
```diff
@@ -0,0 +1,154 @@
...
+  private async downloadStylua(): Promise<void> {
+    const version = util.getDesiredVersion();
+    const release = await this.github.getRelease(version);
+    const assetFilename = util.getAssetFilenamePattern();
+    const outputFilename = util.getDownloadOutputFilename();
+
+    for (const asset of release.assets) {
+      if (assetFilename.test(asset.name)) {
+        const file = createWriteStream(
+          vscode.Uri.joinPath(this.storageDirectory, outputFilename).fsPath,
```
JohnnyMorganz, line 88: Rather than overwriting a single binary each time we download, should we cache each binary, and name them `stylua-vX.Y.Z`? This means we only need to download a binary once, and if a user changes versions depending on their project often, we don't keep starting download requests.

Maybe this is out of scope for this PR though

JohnnyMorganz: Another thing, which is also out of scope for this PR, is probably to write the new extension to a temporary location first (incase someone exits VSCode during a download), then move the fully downloaded binary to the wanted location afterwards, for atomicity

JohnnyMorganz (commented):
This all looks good to me having tested it. Just a few comments for consideration, but nothing major hopefully.

stylua-vscode/src/download.ts
LastTalon: I can change it to what you like, but when I was testing it, jamming it all into one prompt looked quite ugly. Unfortunately VSC doesn't think that notifications deserve formatting, so it removes any newlines or attempt at formatting.

Let me know your thoughts.

stylua-vscode/src/download.ts
LastTalon: I think you're right on all counts. Those are good ideas, but out of scope of this PR I think. Most of this code is simply being moved to encapsulate it.

stylua-vscode/src/download.ts
LastTalon: An alternative we could use is to make it modal which does get formatting, but that may be too intrusive.

JohnnyMorganz: I think a modal is too intrusive here. I think it's fine as it is for now.

Should we bother checking the error for a 429/ratelimit error before showing this specific prompt? Not really sure what other errors could happen

LastTalon: We could do that. My thinking was that if we're getting errors and making more requests there's a risk of rate limit anyway, but if you want we can hold of displaying that until we start hitting the rate limit.

JohnnyMorganz: No I think you're right, this is fine.

JohnnyMorganz (approved):

stylua-vscode/src/download.ts
LastTalon: We can always listen for feedback on it. This is a relatively easy fix if we need to later.

PR #245, opened 2021-08-22
Improve CLI error reporting

JohnnyMorganz:
- Library errors are now moved into a custom enum, allowing us to access specific information from the binary.
- Improve error reporting for "no file or directory found", to make it easier to read.
Old: `error: could not walk: foobar: IO error for operation on foobar: The system cannot find the file specified. (os error 2)`
New: `error: no file or directory found matching 'foobar'`
- Reduced verbosity of parsing errors
- Added colour to error output

TODO maybe:
- Look into making other errors easier to read
- Maybe using codespan diagnostics for syntax error reporting - need to make sure this is still machine readable.
- JSON error output? good for extensions etc.


matthargett (approved):
would be nice to see some test coverage verifying the exception messages, but the color styling changes LGTM otherwise!

JohnnyMorganz:
Indeed, currently unfortunately none of the CLI is tested yet which isn't great. Need to set aside some time to do it

PR #268, opened 2021-09-16
Update extension changelog

LastTalon:
This updates the changelog to include changes from #241. Sorry.


JohnnyMorganz:
Can you mentioning something small about why we need authentication to GitHub/to fix ratelimits? As this wasn't necessary in the previous versions.

I know in the changes section we mentioned about interacting with GitHub API directly already - so maybe something like "Added optional authentication with GitHub to interact with the GitHub REST API"

PR #323, opened 2021-12-24
Switch pre-commit language to Rust so pre-commit can build StyLua from source

rouge8:
The previous pre-commit config required having StyLua already installed. With this patch, pre-commit is responsible for building its own StyLua.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,6 +3,6 @@
   name: StyLua
   description: An opinionated Lua code formatter
   entry: stylua
-  language: system
+  language: rust
   types:
     - lua


JohnnyMorganz:
Hm, is it possible to support both? Some people might have Rust/Cargo installed whilst others just want to download a prebuilt binary and leave it in their PATH

rouge8:
Not that I know of. For people using a pre-built binary without Rust installed, they can always add a local pre-commit hook using the `system` language. For a remote hook like this, another advantage of `language: rust` is being able to use `pre-commit autoupdate` to update and pin the latest version.

JohnnyMorganz:
I've never really used precommit hooks, so not really sure what to do.
I'm going to merge this in though, as the update/version pinning seems like a good feature here, hopefully it doesn't break anything!

ViViDboarder:
It is possible to have both. You can have two hooks in a single repo. One can be of type `rust` and the other of type `system` and call them `stylua` and `stylua-system` or `stylua` and `stylua-cargo`.

The reason the system version is valuable is that some users may be developing Lua but not have cargo installed since the Rust toolchain is quite large and compilation can take a lot of time.

PR #330, opened 2022-01-03
feat: Add more granular control over call parentheses

### LINKED ISSUE #329: feat: Add an option to remove parethesis arround table call
@JohnnyMorganz firstly great work on stylua . Thanks for making it happen.

I like having having function call parenthesis on only string call like
```lua
fn("asdf")
```
But not when only table is used for function call . like
```lua
fn { x = 'a', y = 'b' }
```
Two layers of parenthesis on 
```lua
fn({x='a', y='b'})
```
 seems unnecessary and no parens when only string is used gets weird time to time
```lua
require 'xyz' '32'
```

Currently it's only possible to either have parenthesis on both cases or on none . Would've really liked if there was a middle ground .

Is it possible to add an option to only remove parenthesis on table calls ?

It's farely easy to add
```diff
diff --git a/src/formatters/functions.rs b/src/formatters/functions.rs
index 1a356e9..1101f4f 100644
--- a/src/formatters/functions.rs
+++ b/src/formatters/functions.rs
@@ -107,7 +107,7 @@ pub fn format_function_args(
             arguments,
         } => {
             // Handle config where parentheses are omitted, and there is only one argument
-            if ctx.config().no_call_parentheses
+            if (ctx.config().no_call_parentheses || ctx.config().no_table_call_parentheses)
                 && arguments.len() == 1
                 && !matches!(call_next_node, FunctionCallNextNode::ObscureWithoutParens)
             {
@@ -115,12 +115,16 @@ pub fn format_function_args(
                 if let Expression::Value { value, .. } = argument {
                     match &**value {
                         Value::String(token_reference) => {
-                            return format_function_args(
-                                ctx,
-                                &FunctionArgs::String(token_reference.to_owned()),
-                                shape,
-                                call_next_node,
-                            );
+                            if ctx.config().no_call_parentheses {
+                                return format_function_args(
+                                    ctx,
+                                    &FunctionArgs::String(token_reference.to_owned()),
+                                    shape,
+                                    call_next_node,
+                                );
+                            } else {
+                                ();
+                            }
                         }
                         Value::TableConstructor(table_constructor) => {
                             return format_function_args(
@@ -506,7 +510,7 @@ pub fn format_function_args(
         }
 
         FunctionArgs::TableConstructor(table_constructor) => {
-            if ctx.config().no_call_parentheses
+            if (ctx.config().no_call_parentheses || ctx.config().no_table_call_parentheses)
                 && !matches!(call_next_node, FunctionCallNextNode::ObscureWithoutParens)
             {
                 let table_constructor = format_table_constructor(ctx, table_constructor, shape)
diff --git a/src/lib.rs b/src/lib.rs
index aa1510d..e7b77be 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,8 @@ pub struct Config {
     /// Whether to omit parentheses around function calls which take a single string literal or table.
     /// This is added for adoption reasons only, and is not recommended for new work.
     no_call_parentheses: bool,
+    /// Whether to omit parentheses around function calls which take only a single table.
+    no_table_call_parentheses: bool,
 }
 
 impl Config {
@@ -149,6 +151,13 @@ impl Config {
             ..self
         }
     }
+    /// Returns a new config with the given value for [`no_table_call_parentheses`]
+    pub fn with_no_table_call_parentheses(self, no_table_call_parentheses: bool) -> Self {
+        Self {
+            no_table_call_parentheses,
+            ..self
+        }
+    }
 }
 
 impl Default for Config {
@@ -160,6 +169,7 @@ impl Default for Config {
             indent_width: 4,
             quote_style: QuoteStyle::default(),
             no_call_parentheses: false,
+            no_table_call_parentheses: false,
         }
     }
 }
```

If you're in favor of the idea I can open a pr .

shadmansaleh:
Adds new option call_parentheses with possible values Always, NoTable,
     None (Default: Always)

When call_parentheses is set to Always stylua applies call parentheses
all the time.
```lua
xyz({ a = 1, b = 2 })
abc('Hello')
```
When set to NoTable it omits parentheses on function call with
single table argument.
```lua
xyz { a = 1, b = 2 }
abc('Hello')
```
And when it's None stylua omits parentheses on function call with
single table or string argument. This is same as `no_call_parentheses =
true`
```lua
xyz { a = 1, b = 2 }
abc 'Hello'
```

closes #329


clason:
For the sake of symmetry, it'd be nice to allow for the converse option (`NoString`), too.

And personally I'd flip it to have `TableOnly` and `StringOnly` (opt-in instead of opt-out) ;)

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Added option `call_parentheses`
+Specify whether to apply parentheses on function calls with single string or table arg. Possible options:`Always`, `NoTable`, `None`.
+
+### Deprecated
+- Option `no_call_parentheses` has been deprecated. Use no_call_parentheses = "None" instead.
```
JohnnyMorganz, line 13:
```suggestion
- Option `no_call_parentheses` has been deprecated. Use `call_parentheses = "None"` instead.
```

README.md
```diff
@@ -135,7 +135,8 @@ StyLua only offers the following options:
...
-| `no_call_parentheses` | `false` | A style option added for adoption purposes. When enabled, parentheses are removed around function arguments where a single string literal/table is passed. Note: parentheses are still kept in some situations if removing them will make the syntax become obscure (e.g. `foo "bar".setup -> foo("bar").setup`, as we are indexing the call result, not the string)
+| `no_call_parentheses` (deprecated) | `false` | A style option added for adoption purposes. When enabled, parentheses are removed around function arguments where a single string literal/table is passed. Note: parentheses are still kept in some situations if removing them will make the syntax become obscure (e.g. `foo "bar".setup -> foo("bar").setup`, as we are indexing the call result, not the string). Note: this option has been deprecated use call_parentheses = "None" instead.
```
JohnnyMorganz, line 138: I would probably just remove this option from the README completely, as it is no longer relevant for new configs

README.md
```diff
@@ -145,4 +146,5 @@ indent_type = "Tabs"
 indent_width = 4
 quote_style = "AutoPreferDouble"
 no_call_parentheses = false
```
JohnnyMorganz, line 148: Could also remove this from the list
```suggestion
```

tests/test_call_parentheses.rs
```diff
@@ -0,0 +1,244 @@
+use stylua_lib::{format_code, CallParenType, Config, OutputVerification};
+
+fn format(paren_type_str: &str, input: &str) -> String {
+    let paren_type = match paren_type_str {
+        "Always" => CallParenType::Always,
+        "NoTable" => CallParenType::NoTable,
+        "None" => CallParenType::None,
+        _ => CallParenType::Always,
+    };
```
JohnnyMorganz, line 9: Rather than matching a string, could we just pass in the enum variant directly to the function? Gives us an immediate compiler warning and reduces the likelihood of typos etc.

src/formatters/functions.rs
```diff
@@ -464,7 +470,8 @@ pub fn format_function_args(
         }
 
         FunctionArgs::String(token_reference) => {
-            if ctx.config().no_call_parentheses
+            if (ctx.config().no_call_parentheses
+                || ctx.config().call_parentheses == CallParenType::None)
```
JohnnyMorganz, line 474: It might be nice to abstract out these condition checks onto a method implemented on config, so instead we can just call, e.g. something like
`ctx.config().should_omit_string_parens()` -> no_call_parentheses = true || CallParenType == None
`ctx.config().should_omit_table_parens()` -> no_call_parentheses = true || CallParenType == NoTable || CallParenType == None

Would help in future if we ever decide to add e.g. NoString or something

README.md
```diff
@@ -135,7 +135,8 @@ StyLua only offers the following options:
...
+| `no_call_parentheses` (deprecated) | `false` | A style option added for adoption purposes. When enabled, parentheses are removed around function arguments where a single string literal/table is passed. Note: parentheses are still kept in some situations if removing them will make the syntax become obscure (e.g. `foo "bar".setup -> foo("bar").setup`, as we are indexing the call result, not the string). Note: this option has been deprecated use call_parentheses = "None" instead.
+| `call_parentheses` | `Always` | Specify whether to apply parentheses on function calls with single string or table arg. Possible options:`Always`, `NoTable`, `None`].When call_parentheses is set to Always stylua applies call parentheses all the time. When set to NoTable it omits parentheses on function call with single table argument. And when it's None stylua omits parentheses on function call with single table or string argument (same as no_call_parentheses).
```
JohnnyMorganz, line 139:
```suggestion
| `call_parentheses` | `Always` | Specify whether to apply parentheses on function calls with a single string or table argument. Possible options: [`Always`, `NoTable`, `None`]. When `call_parentheses` is set to `Always`, StyLua applies call parentheses all the time. When set to `NoTable` it omits parentheses on function calls with a single table argument. And when it's `None` StyLua omits parentheses on function call with single table or string argument (originally as `no_call_parentheses`). Note: parentheses are still kept in some situations if removing them will make the syntax become obscure (e.g. `foo "bar".setup -> foo("bar").setup`, as we are indexing the call result, not the string).
```

JohnnyMorganz (changes requested):

shadmansaleh:
@JohnnyMorganz let me know what ypu think about clasons [suggestions](https://github.com/JohnnyMorganz/StyLua/pull/330#issuecomment-1004049189)

JohnnyMorganz:
Ok, I think adding an option for strings is sensible as well for the sake of consistency, it isn't really adding anything major either

Technically these are opt-outs from the default (Always), so I think `NoString`/`NoTable` are more appropriate, although `StringOnly`/`TableOnly` kinda do sound nicer. Not sure

shadmansaleh:
I've added `NoString`

tests/test_call_parentheses.rs
```diff
@@ -0,0 +1,298 @@
+use stylua_lib::{format_code, CallParenType, Config, OutputVerification};
```
JohnnyMorganz, line 1: Is this similar to `test_no_call_parens.rs`? Should we just delete the no_call_parens file and move any test cases over there into here, with `CallParenType::None`.

shadmansaleh: It's pretty similar . I can move the tests . But having the no_call_parens tests also insures currently deprecated no_call_paren option doesn't get accidentally broken (I'm not sure if we care about this)

JohnnyMorganz: Hm, true - leave it for now. When the option gets removed I can move over any new tests

JohnnyMorganz (approved):
Everything looks good to me, the only thing I'm not sure of is whether to go with `NoString`/`NoTable` or `TableOnly`/`StringOnly`

shadmansaleh:
Let me add one more to that confusion . How about `NoSingleString`, `NoSingleTable` . I think that conveys the correct message about what the option does . NoString or TableOnly sounds like parens will be dropped strimg args always .

JohnnyMorganz:
I agree, that sounds better. Lets go with `NoSingleString` / `NoSingleTable` - I think the verbosity of the name is reasonable

JohnnyMorganz:
Thank you!

clason:
@JohnnyMorganz what do you think about a new release? With this change, we could consider enforcing StyLua on Neovim's codebase :)

JohnnyMorganz:
> @JohnnyMorganz what do you think about a new release? With this change, we could consider enforcing StyLua on Neovim's codebase :)

I'm just waiting on a new full-moon release upstream, hopefully by the end of the week. Once that's out, I'll cut a new stylua release

PR #337, opened 2022-01-07
Multiline formatting of generic for syntax

### LINKED ISSUE #322: Total line length not taken into account in for loops
I have a for loop with a long initial line. StyLua formats it like this:

```lua
local function system(world)
	for id, model, lasering, transform in world:query(Components.Model, Components.Lasering, Components.Transform, Components.Mothership) do
	end
end
```

If I make the function call longer, StyLua formats it more correctly:

```lua
local function system(world)
	for id, model, lasering, transform in world:query(
		Components.Model,
		Components.Lasering,
		Components.Transform,
		Components.Mothership,
		Components.Mothership
	) do
	end
end
```

But if I add more variables to the loop, it does not do anything:

```lua
local function system(world)
	for id, model, lasering, transform, id, model, lasering, transform, id, model, lasering, transform, id, model, lasering, transform in world:query(Components.Model, Components.Lasering, Components.Transform, Components.Mothership) do
	end
end
```

Ideally, the formatter should probably produce something close to this:

```lua
local function system(world)
	for id, model, lasering, transform,
		id, model, lasering, transform, 
		id, model, lasering, transform, 
		id, model, lasering, transform 
	in
		world:query(
			Components.Model,
			Components.Lasering,
			Components.Transform,
			Components.Mothership,
			Components.Model,
			Components.Lasering,
			Components.Transform,
			Components.Mothership
		)
	do
	end
end
```

JohnnyMorganz:
Closes #322

### DIFF
--- a/tests/snapshots/tests__standard@generic-for-multiline.lua.snap
+++ b/tests/snapshots/tests__standard@generic-for-multiline.lua.snap
@@ -1,0 +1,63 @@
+---
+source: tests/tests.rs
+expression: format(&contents)
+
+---
+local function system(world)
+	for id, model, lasering, transform in world:query(
+		Components.Model,
+		Components.Lasering,
+		Components.Transform,
+		Components.Mothership
+	) do
+	end
+end
+
+local function system(world)
+	for id,
+		model,
+		lasering,
+		transform,
+		id,
+		model,
+		lasering,
+		transform,
+		id,
+		model,
+		lasering,
+		transform,
+		id,
+		model,
+		lasering,
+		transform
+	in world:query(Components.Model, Components.Lasering, Components.Transform, Components.Mothership) do
+	end
+end
+
+local function system(world)
+	for id,
+		model,
+		lasering,
+		transform,
+		id,
+		model,
+		lasering,
+		transform,
+		id,
+		model,
+		lasering,
+		transform,
+		id,
+		model,
+		lasering,
+		transform
+	in world:query(
+		Components.Model,
+		Components.Lasering,
+		Components.Transform,
+		Components.Mothership,
+		Components.Mothership
+	) do
+	end
+end
+
--- a/tests/snapshots/tests__standard@generic-for-multiline.lua.snap
+++ b/tests/snapshots/tests__standard@generic-for-multiline.lua.snap
@@ -4,17 +4,15 @@ expression: format(&contents)
 
 ---
 local function system(world)
-	for id, model, lasering, transform in world:query(
-		Components.Model,
-		Components.Lasering,
-		Components.Transform,
-		Components.Mothership
-	) do
+	for id, model, lasering, transform in
+		world:query(Components.Model, Components.Lasering, Components.Transform, Components.Mothership)
+	do
 	end
 end
 
 local function system(world)
-	for id,
+	for
+		id,
 		model,
 		lasering,
 		transform,
@@ -35,7 +33,8 @@ local function system(world)
 end
 
 local function system(world)
-	for id,
+	for
+		id,
 		model,
 		lasering,
 		transform,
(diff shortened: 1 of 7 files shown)


tests/snapshots/tests__standard@generic-for-multiline.lua.snap
```diff
@@ -0,0 +1,63 @@
...
+		model,
+		lasering,
+		transform
+	in world:query(
+		Components.Model,
+		Components.Lasering,
+		Components.Transform,
+		Components.Mothership,
+		Components.Mothership
+	) do
```
matthargett, line 60: hm. Is there a reason we don't align the `do` with the `end`, like we do for similar scope closure pairs like: `function()` and `end`, `if` and `else`?

tests/snapshots/tests__standard@generic-for-multiline.lua.snap
```diff
@@ -0,0 +1,63 @@
...
+		id,
+		model,
+		lasering,
+		transform
+	in world:query(Components.Model, Components.Lasering, Components.Transform, Components.Mothership) do
+	end
+end
+
+local function system(world)
+	for id,
```
matthargett, line 38: why isn't `id` indented on the next line like the rest of the arguments? again, I'm thinking of consistency with the `if` and `else` cases

tests/snapshots/tests__standard@generic-for-multiline.lua.snap
JohnnyMorganz: In this case, I followed [how prettier did it](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAzTBeTACgEp8A+TYAHSkxwgCcjIoNMBtASwBMAaTALYRucADb9RAQzRwGnKAHN+MBpNbZGAgLqYIuAO6NR3AHQBHAK6yAnoQDCEAQAdoCGGhMBZYWP4PnrrAeADLSsvJKmP4uUG4eACqq6pp+jjFxXhAwABayaNmcTsSk1LR0mCxoEKJwJqIQCoRUIPAYzcQ0dAC+ND1QNJy4hIYMxuZWDLbRge6ZIuJRaTMhYXKKqQGxQSaJamgaDAIb6dveOXkFTsfLmecM+YXXW7NnufeXxZSdmL0gvCAQJwwTjoZCgSQMBgQfQABQhCDQyBAklE+kk1kR-wARqowABrOAwADKkgEcGC8jgyGwKJk2NxBOJTkkYAiyBUVn+cAEWLg3BE3FCigskgUcAAYppJDBgYokZILDAIH8QNkYAJRAB1AqtZlgOBEhGcYEAN2N1iRYDQmJA8hkDBgMNUCgEkmptLg-wAVmgAB5EiI1ACKFiyVKQNNEdJAzPusiRWMkvNEKqcaxgmp4OWQAEYAEwABn+aYgMk1qicSLTcHtJqp-0sYadgMRKGkAFpYny+SqGHBLJw+07Ra73VHPSAZAJOOyGJzJ4G4CGw2PozAk5nuNmkHn-ipJJxRBF-KOUDWAKwqiwyeJJ1uR6MmqwASSgIlgRLAciBAEE30SYGsGpVzgLouiAAl)

It seems that in a for loop, the body opening bracket is aligned with the closing parentheses, but in an if-statement they don't do this...

I think I agree, its probably better to just keep it consistent with how we do `if ... then` expanded multiline. Also makes it easier to handle comments

tests/snapshots/tests__standard@generic-for-multiline.lua.snap
JohnnyMorganz: I don't think there was a particular reason for this, just how I did it (its interesting how the the `id` in the first line aligns nicely with the other variables too).

Again, I agree its better to be consistent here so I'll change it

JohnnyMorganz:
@matthargett, wanted your opinion on the changes (I can't seem to re-request your review)

matthargett (approved):

matthargett:
I didn’t test locally, but reviewing the tests it looks good to me.

PR #342, opened 2022-01-20
Hang assignment at equal token before expanding RHS

### LINKED ISSUE #292: Assignment: hang at equal token before expanding RHS
Currently I have a function that gets formatted into this:
```lua
local musicId, musicTime, responseTick, responseOffset = remotes.Server.GetSpectatorInfo:InvokeServer(
	player,
	sendTick
)
```
In this situation I believe that this:
```lua
local musicId, musicTime, responseTick, responseOffset = 
	remotes.Server.GetSpectatorInfo:InvokeServer(player, sendTick)
```
Or maybe even this: (This cuts down on variables per a line which maybe is be easier to read)
```lua
local musicId, musicTime, 
	responseTick, responseOffset = remotes.Server.GetSpectatorInfo:InvokeServer(player,sendTick)
```
Looks better.

JohnnyMorganz:
Closes #292 

TODO:
- are we happy with the test changes?
- Update changelog


tests/snapshots/tests__standard@large-example-2.lua.snap
```diff
@@ -1552,20 +1552,8 @@ script.Parent.Parent.Humanoid.Health = 0
...
-							1,
-							0,
-							-1,
-							0,
-							1,
-							0,
-							0
-						) --math.random(),math.random(),math.random(),math.random(),math.random(),math.random(),math.random(),math.random(),math.random()) -- i like the people being upside down better.
+						torso.CFrame =
+							CFrame.new(torso.Position.x, torso.Position.y, torso.Position.z, 0, 0, 1, 0, -1, 0, 1, 0, 0) --math.random(),math.random(),math.random(),math.random(),math.random(),math.random(),math.random(),math.random(),math.random()) -- i like the people being upside down better.
```
matthargett, line 1556: It doesn’t still break the line budget this way? It seems like when it breaks at the assignment, and that doesn’t fit within the budget, then it should fall back to the other approaches.

JohnnyMorganz: Yep, this line is at exactly 120 chars, so is technically correct! (We exclude comments from the count)

JohnnyMorganz:
@matthargett what are your thoughts on the diffs of the other two test cases which I've just committed? I'm not sure I'm happy about them

matthargett:
Hm. I wonder if hanging at interstitial operator starting at the end of the line and working backward has a better outcome? This is a step in the right direction, regardless. I feel like what we are mostly seeing here is the reality of 120 character line lengths and *consistent* application of styling rules.

PR #344, opened 2022-01-24
Fix call_parentheses

warfurret:
`call_parentheses` should be `no_call_parentheses` as `call_parentheses` is not a valid config in the stylua.toml file.


warfurret:
I actually just found out that what's happening is the `stylua-action` currently doesn't have `call_parentheses`. Disregard.

JohnnyMorganz:
Apologies, this was recently changed but it hasn't been released yet. The README shows latest changes

PR #350, opened 2022-01-31
Fix misformatting of if-expression syntax

### LINKED ISSUE #349: if-expressions can lose indentation upon reformatting
A few examples of this in 0.12:
```lua
				local function append(cache: ApolloCache<any>, item: Item)
					local data = cache:readQuery({ query = query })
					cache:writeQuery({
						query = query,
						data = Object.assign({}, data, {
							items = Array.concat({}, (function()
								local ref = if Boing.toXYZBoxinf(data) and data ~= nil
										then  data.items
										else data
								return Boing.toXYZBoxinf(ref) and ref
							end)() or {}, { item }),
						}),
					})
					return item
				end
```

```lua
	local error_ = if errors and #(errors :: Array<any>) > 0
		then ApolloError.new({ graphQLErrors = errors })
		else nil
```

```lua
local function useMutation<TData, TVariables, TContext, TCache>(
	mutation: DocumentNode | TypedDocumentNode<TData, TContext>,
	options: MutationHookOptions_<TData, TVariables, TContext>?
): MutationTuple<TData, TVariables, TContext, TCache>
	local context = useContext(getApolloContext())
	local result, setResult = useState({ called = false, loading = false })
	local updatedOptions = if options
		then Object.assign({}, options, { mutation = mutation })
		else { mutation = mutation }
```

JohnnyMorganz:
Fixes #349

### DIFF
--- a/src/formatters/expression.rs
+++ b/src/formatters/expression.rs
@@ -454,7 +454,7 @@ fn format_if_expression(ctx: &Context, if_expression: &IfExpression, shape: Shap
     );
 
     // Determine if we need to hang the expression
-    let singleline_shape = (shape + 3 + 5) // 3 = "if " + 5 = " then"
+    let singleline_shape = (shape + 3 + 6 + 6) // 3 = "if " + 6 = " then " + 6 " else "
         .take_first_line(&strip_trivia(&singleline_condition))
         .take_first_line(&strip_trivia(&singleline_expression))
         .take_first_line(&else_ifs.as_ref().map_or(String::new(), |x| {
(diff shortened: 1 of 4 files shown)


src/formatters/expression.rs
```diff
@@ -454,7 +454,7 @@ fn format_if_expression(ctx: &Context, if_expression: &IfExpression, shape: Shap
     );
 
     // Determine if we need to hang the expression
-    let singleline_shape = (shape + 3 + 5) // 3 = "if " + 5 = " then"
+    let singleline_shape = (shape + 3 + 6 + 6) // 3 = "if " + 6 = " then " + 6 " else "
```
matthargett, line 457: can you extract `6` here to an explaining variable? like `local thenOrElseLength = 6`

matthargett (approved):

PR #359, opened 2022-02-04
Fix crash from unhandled singleton types

### LINKED ISSUE #358: Crash with Singleton Types
`Stylua v0.12.1` is crashing when declaring a new type with singleton types

Code:
```lua
local a: "foo" | "bar" = "foo" -- OK
type B = "foo" -- OK
type C = "foo" | "bar" -- NOT OK
``` 

That last type declaration crashes causing this error:
```bash
thread '<unnamed>' panicked at 'unknown node String(TokenReference { leading_trivia: [],
token: Token { start_position: Position { bytes: 55, line: 3, character: 10 },
end_position: Position { bytes: 60, line: 3, character: 15 },
token_type: StringLiteral { literal: ShortString("foo"),
multi_line: None, quote_type: Double } },
trailing_trivia: [Token { start_position: Position { bytes: 60, line: 3, character: 15 },
end_position: Position { bytes: 61, line: 3, character: 16 },
token_type: Whitespace { characters: ShortString(" ") } }] })',
/stylua-0.12.1/src/formatters/trivia_util.rs:187:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

JohnnyMorganz:
Also updates the full moon test cases to latest
Fixes #358


matthargett:
Verified that this fixes the crashes reported by @flakolefluk and I reproduced across a few other non-trivial files.

matthargett (approved):

PR #363, opened 2022-02-07
Add Sublime Text Package to Other Installation Methods

aerobounce:
This adds an link item to Lua formatter plugin for Sublime Text I made, which wraps `StyLua` and ships with some additional features.
Hope this is the correct section to add.


JohnnyMorganz:
Looks useful, thanks for making it!

PR #402, opened 2022-03-07
Hang type union declarations if multiline comment is seen

JohnnyMorganz:


matthargett:
Might be worth capturing the intersection cases as well

PR #408, opened 2022-03-16
Break down function argument code

JohnnyMorganz:
Cleaning up the tech debt, this code was very difficult to read.
Its still not great, but its an improvement


JohnnyMorganz:
Why are actions not running 🤔

PR #424, opened 2022-03-28
Fix comments ordering when semicolon removed after stmt

### LINKED ISSUE #423: Comments order is changed for statements with multiple comments and a semicolon at the end
### Input code

```lua
local foo = a <= b --[[ some block comment ]]; -- inline comment
fn() --[[ some block comment 2 ]]; -- inline comment 2
```

### Current output

```lua
local foo = a <= b -- inline comment --[[ some block comment ]]
fn() -- inline comment 2 --[[ some block comment 2 ]]
```

### Expected output

```lua
local foo = a <= b --[[ some block comment ]] -- inline comment
fn() --[[ some block comment 2 ]] -- inline comment 2
```

JohnnyMorganz:
Fixes #423


JohnnyMorganz:
Uncovered lines are impossible to cover, so it is a non-issue

Issue #430, opened 2022-04-01
It requires 2 runs of Stylua so stabilize the output for some examples


ktrz:
All of the examples are run wth the following configuration:

StyLua version: `13.0.0`

`.stylua.toml` file:
```toml
column_width = 100
line_endings = "unix"
indent_type = "T=tabs"
indent_width = 4
quote_style = "double"
```

### The following examples require 2 runs of StyLua to produce proper (and stable) results

```lua
type Foo<T = --[[leading]]
string
--[[trailing]]> = { baz: T, }
```

```lua
type Bar<T
--[[ Trailing comment ]]> = {}
```

```lua
-- This is a comment before
type Foo = --[[ Comment before Bar ]]
Bar<--[[ Before X ]]
X, --[[ After X ]]
--[[ Before Y ]]
Y, --[[ After Y ]]
--[[ Before Z ]]
Z
--[[ After Z ]]> -- This is a comment after
```

```lua
type OffsetEnum = { [string]: number, }
local OffsetEnum = {aValue = 10, anotherValue = 11, yetAnotherValue = 12, reset = 0, postReset = 1}
```

### Results after the first run

```lua
type Foo<
	T =  --[[leading]]string	--[[trailing]]

> = { baz: T }

----------------

type Bar<T--[[ Trailing comment ]]
> = {}

-----------------

-- This is a comment before
type Foo =  --[[ Comment before Bar ]]Bar<--[[ Before X ]]X,  --[[ After X ]]--[[ Before Y ]]
Y,  --[[ After Y ]]--[[ Before Z ]]
Z--[[ After Z ]]
> -- This is a comment after

-------------------

type OffsetEnum = { [string]: number }
local OffsetEnum =
	{
		aValue = 10,
		anotherValue = 11,
		yetAnotherValue = 12,
		reset = 0,
		postReset = 1,
	}
```

### Results after the second and any consecutive run

```lua
type Foo<
	T =  --[[leading]]string --[[trailing]]
> = { baz: T }

----------------

type Bar<
	T--[[ Trailing comment ]]
> = {}

-----------------

-- This is a comment before
type Foo =  --[[ Comment before Bar ]]Bar<--[[ Before X ]]X,  --[[ After X ]]--[[ Before Y ]]Y,  --[[ After Y ]]--[[ Before Z ]]Z--[[ After Z ]]> -- This is a comment after

-------------------

type OffsetEnum = { [string]: number }
local OffsetEnum = {
[...]

PR #435, opened 2022-04-02
Fix unstable formatting around comments in type declarations

### LINKED ISSUE #397: Inconsistent formatting when running stylua multiple times
given:
```
--[[comment]]
type Doo
--[[comment]]
<
--[[comment]]
T
--[[comment]]
>
--[[comment]]
= 
--[[comment]]
string
--[[comment]]
```

after stylua becomes:
```
--[[comment]]
type Doo--[[comment]]
<
	--[[comment]]
	T	--[[comment]]

>--[[comment]]
 = --[[comment]]
string
--[[comment]]
```
and running it a second time it becomes:
```
--[[comment]]
type Doo--[[comment]]<
	--[[comment]]
	T --[[comment]]
>--[[comment]] =  --[[comment]]string
--[[comment]]
```
running it more times stays the same.

Expected result: 
* results should be the same everytime stylua is executed

Tested on v0.12.4

JohnnyMorganz:
Fixes #397 (ref: #430)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed leading trivia on semicolon lost when semicolon is removed ([#431](https://github.com/JohnnyMorganz/StyLua/issues/431))
 - Fixed shape calculation of the RHS of a binary expression not correctly reset when hanging, causing it to expand unnecessarily ([#432](https://github.com/JohnnyMorganz/StyLua/issues/432))
+- Fixed unstable formatting around comments within type declarations ([#397](https://github.com/JohnnyMorganz/StyLua/issues/397), [#430](https://github.com/JohnnyMorganz/StyLua/issues/430))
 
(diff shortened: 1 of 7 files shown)


JohnnyMorganz:
@matthargett @flakolefluk @ktrz thoughts on the new snapshots?

The newly added snapshots of your tests looks good, but I'm not 100% on the change in the table snapshot (its happening because there is a comment in between the equals token and the table in the declaration)

matthargett:
innnnnteresting. it would be nice if it didn't force the open brace onto the next line with an indent, but I also think the trade-off here is a good one since the affected regression case is a weird/rare one.

matthargett (approved):

ktrz (approved):
I think the change in this one snapshot is a good tradeoff for having the stable tool output.

flakolefluk:
snapshots look good to me! i agree with  the previous comments. Let's start with making the output stable.  🎉

JohnnyMorganz:
Turns out the above diff is not related to the PR, but is actually a regression from 0.12.5 -> 0.13.0

matthargett (approved):

PR #444, opened 2022-04-16
Statically link linux build

stopdropandrew:
stylua's linux build fails to run on Ubuntu 18:
```
./stylua: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by ./stylua)
```

One oddity was that `--target` must be passed explicitly, as documented by this [blog post](https://msfjarvis.dev/posts/building-static-rust-binaries-for-linux/)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -11,6 +11,3 @@ rustflags = [
     "-C",
     "link-args=/NODEFAULTLIB:libucrt.lib /NODEFAULTLIB:libucrtd.lib /NODEFAULTLIB:ucrtd.lib",
 ]
-
-[target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "target-feature=+crt-static"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -11,3 +11,6 @@ rustflags = [
     "-C",
     "link-args=/NODEFAULTLIB:libucrt.lib /NODEFAULTLIB:libucrtd.lib /NODEFAULTLIB:ucrtd.lib",
 ]
+
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-C", "target-feature=+crt-static"]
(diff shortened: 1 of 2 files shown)


JohnnyMorganz:
Thanks for the PR!

Could we instead put this flag in the `.cargo/config.toml` file similar to this? https://github.com/rojo-rbx/rojo/blob/master/.cargo/config.toml
Hopefully that still should work (and maybe the target won't need to be specified anymore in the build command either?)

stopdropandrew:
I did that initially, seen here https://github.com/JohnnyMorganz/StyLua/runs/6049194259?check_suite_focus=true and other action failures, but I was getting the same error mentioned in that blog post.

I thought the easiest solution was to statically link **only** the release build, rather than figuring out how to pass in `--target` to every command - it's not even clear that's possible. Do you have a preference?

JohnnyMorganz:
Nope, no particular preference here!

Did a bit more digging into the actual cause of this issue (wasn't happening before), and stumbled across rust-analyzer having the same issue: https://github.com/rust-lang/rust-analyzer/issues/11558

It turns out this is because they switched their CI to Ubuntu 20.04 from 18.04, and Rust now requires a newer version of glibc. Our action is on `ubuntu-latest` which defaults to 20.04.

Maybe instead of statically linking the whole C runtime here, we just downgrade our CI to Ubuntu 18.04? I think that should also fix the problem, and I don't see any downside to it

LastTalon:
> Maybe instead of statically linking the whole C runtime here, we just downgrade our CI to Ubuntu 18.04? I think that should also fix the problem, and I don't see any downside to it

I think this is the way to go. The issue is with incompatible ABI changes. The newer version will be able to handle the old version linking, just not the other way around.

Possibly in the future (depending on support) CI could even build for each version of libc? This may be unsupported by foreman workflows though.

PR #445, opened 2022-04-18
Build release on Ubuntu 18

stopdropandrew:
Changed all runners to use `ubuntu-18.04`.

I closed https://github.com/JohnnyMorganz/StyLua/pull/444 in favor of building on ubuntu 18 and older version of glibc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
     - test-actions
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


matthargett:
will these binaries be compatible with new Ubuntu LTS when it's released?

stopdropandrew:
My understanding is that glibc is generally backwards compatible, but never forwards compatible.

JohnnyMorganz:
Thanks!

PR #451, opened 2022-05-01
Fix some typos

mhmdanas:


src/formatters/expression.rs
```diff
@@ -874,20 +874,20 @@ impl ToRange for Expression {
 /// It holds the range of the leftmost binary expression, and the original additional indent level of this range.
 /// This struct is only used when the hanging binary expression involves a hang level, for example:
 /// ```lua
-/// foooo
```
mhmdanas, line 877: Is this intentional? It feels intentional, but I'm not sure.

JohnnyMorganz: Can't even remember, the correction to `foo` looks fine to me though in this case

JohnnyMorganz (approved):
Thanks!

PR #454, opened 2022-05-03
Sort requires codemod

### LINKED ISSUE #364: Sort requires
It would be really neat if stylua could re-order my requires in alphabetical order.

It should be an optional parameter because stylua can't assume that require does not have side effects, but typically it's good practice to avoid them, so generally it's fine 🙂

JohnnyMorganz:
An implementation of the sort requires codemod

Enabling sort requires functionality will make stylua sort requires in the top level block of a file accordingly.
All requires will move to the top, and will be ordered alphabetically.
Any dependencies used within requires (e.g. `ReplicatedStorage` in `require(ReplicatedStorage.Module)`) will also be moved so that they are sorted above requires.

Assumptions:
- We only touch requires in the form `local NAME = require(EXPRESSION)`. Other types of requires are left alone
- We only touch requires in the top level of a file. We don't touch requires in nested blocks.
- Requires have no side-effects, reorganising requires is safe

TODO:
- Topologically sort dependencies and requires so that dependencies are correctly structured. E.g. `local c = script.Parent; local a = child.Module`, we need to enforce `c` goes before `a`.
- Add ability to structure requires (and other variables) into sections. This should open up functionality to sort `game:GetService()` calls at the top, before requires.
- Add sections ordering.
- Put newlines in appropriate positions
- More tests!

What the configuration file will look like when this is implemented:
```toml
column_width = 120
line_endings = "Unix"
indent_type = "Tabs"
indent_width = 4
call_parentheses = "Always"

[sort_requires]
enabled = true
sections_order = ["services", "packages", "shared", "requires"] # The order to apply sections in. "requires" is the default section when no other section matches

[sort_requires.sections]
# A map between section names and a list of regex's to match against values
services = ["game:GetService(.)"]
packages = ["require(ReplicatedStorage.Packages.([%w.]+))"]
shared = ["require(ReplicatedStorage.Shared.([%w.]+))"]
```

Closes #364

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ use serde::Deserialize;
 mod context;
 mod formatters;
 mod shape;
+mod sort_requires;
 mod verify_ast;
 
 /// The type of indents to use when indenting
@@ -283,6 +284,13 @@ pub fn format_code(
         Some(input_ast.to_owned())
     } else {
         None
+    };
+
+    // If we want to sort requires, we should firstly apply this on the input ast
+    // TODO: connect to option
+    let input_ast = match true {
+        true => sort_requires::sort_requires(input_ast),
+        false => input_ast,
     };
 
     let code_formatter = formatters::CodeFormatter::new(config, range);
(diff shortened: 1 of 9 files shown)


JohnnyMorganz:
Debating whether this really belongs in this tool. It may make more sense for it to be split out into a different tool (similar to isort)

Issue #123, opened 2021-04-17
Is the function call hanging the best possible format?


JohnnyMorganz:
The function call hanging implemented in v0.7.0 helps to improve readability for longer chains, however, some places don't look as great.

We follow the heuristic that we only hang method calls, not static functions. i.e., StyLua would hang as such:
```lua
Promise.new(...)
    :andThen(...)
    :catch(...)
```

However, some APIs don't use a static `.new()` method to create a new object, and may use a `:new()` instead. But StyLua would format it as such:
```lua
FooBar
    :new(...)
    :process(...)
```
But, would it be better for it to be like below?
```lua
FooBar:new(...)
    :process(...)
```
If we were to do this, what kind of heuristic would we have to apply?
Prettier also has a [similar problem](https://youtu.be/0Q4kUNx85_4?t=967). The heuristic they implement is to determine if the first identifier (the prefix in the AST) is uppercase. (e.g. `Object.keys()` and `Promise.new()` vs. `items.filter()`)

A similar heuristic is also implemented for JavaScript `this`. Should we do the same here for `self`?
```lua
self
    :GetOwnedVehicles()
    :andThen(...)
    :catch(...)
```
vs.
```lua
self:GetOwnedVehicles()
    :andThen(...)
    :catch(...)
```

Are there any other places where it looks weird?

LastTalon:
At first glance I would simply say never to hang the first call in a sequence of function calls based purely on non-semantic differences. I think that makes it flow and read appropriately in all of these cases.

I can't think of many places I would desire to break that rule. Even for long function or variable names with the first call I don't think I would want it hanging.

```lua
aVeryLongVariableName.withAVeryExtremelyLongNoGoodFunctionCallName()
```
still seems better to me than
```lua
aVeryLongVariableName
    .withAVeryExtremelyLongNoGoodFunctionCallName()
```
when its the first function being called as a member of that variable.

The clarity of hanging the function calls comes from the them being sequenced similar to items in a multi-line table or function call. So the benefit of hanging the first really only exists in the context of multiple subsequent function calls in my opinion.

```lua
self:first()
    :second()
    :third()
```
is useful for the same reason as
```lua
self = {
    "first",
    "second",
    "third",
}
```
and likewise the usefulness of this formatting is sort of proportional to the length of the list. So for the first term hanging, the benefit is sort of negligible.

That I think allows us to focus more on the semantic aspect of when to hang them or not. This is where I think the core of the issue is.

[...]

PR #470, opened 2022-06-25
Hang static chained function calls

### LINKED ISSUE #368: Function calls not hanging when over column width
For the input below, I expected that the formatted output kept the same amount of lines:

```lua
foo.bar('-------------------------------------------------------------------------------------------------------------')
	.returns()
foo.bar('--------------------------------------------------------------------------------------------------')
	.returns()
	.way()
	.longer()
	.chain()
```

However, StyLua formatted it like the output below, where each function call chain is joined on a line. The lines are longer than the column limit of 120, and that doesn't read nice.

```lua
foo.bar("-------------------------------------------------------------------------------------------------------------").returns()
foo.bar("--------------------------------------------------------------------------------------------------").returns().way().longer().chain()
```

I'm posting it here, because it looks like a duplicate of this issue.

__Originally posted by @TjeuKayim in https://github.com/JohnnyMorganz/StyLua/issues/78#issuecomment-1033593230__

JohnnyMorganz:
Fixes #368


JohnnyMorganz:
Not sure if I'm 100% happy with the changes in the snapshot tests. Suprised the large scale test created no diff though, which is nice

JohnnyMorganz:
Brought up in conversation, maybe we should keep the first call in a chain inlined if it has "simple" or no arguments?
So the .new would be brought back up so its `Promise.new()` and then the chains. Prettier I believe does something similar.

The reason why it must be "simple" is we want to keep indentation aligned, otherwise you get something like this: https://github.com/JohnnyMorganz/StyLua/issues/123#issuecomment-821883893

JohnnyMorganz:
Function definitions unnecessarily hanging at equals token:
image

PR #472, opened 2022-06-25
Fix syntax error field binexp w/ comment collapsed

### LINKED ISSUE #471: Field value with comment binexp collapsed leading to malformed formatting
```lua
local foo = {
	x = props.Item.Type == "Crystal"
		and utf8.char(0x221e) -- Infinite symbol
		or nil,
}
```
... gets formatted as ...
```lua
local foo = {
	x = props.Item.Type == "Crystal" and utf8.char(0x221e) -- Infinite symbol or nil,
}
```
Source: https://github.com/Kampfkarren/zombie-strike/blob/master/src/shared/ReplicatedStorage/Core/UI/Components/ItemDetails.lua

JohnnyMorganz:
Fixes #471


JohnnyMorganz:
Coverage decrease is because all the lines removed were hits, surprised that shows up as negative in codecov... 🤷

PR #476, opened 2022-06-26
Inline first call in chain depending on heuristics

JohnnyMorganz:
Heuristics are same as prettier:
- Prefix begins with uppercase letter, OR
- Prefix is smaller in length than the indent with
AND inlining doesn't go over width


JohnnyMorganz:
In conversation - this diff is not necessary and shouldn't reformat:
image

PR #478, opened 2022-06-26
Fix severe perf regression of complex nested function calls

### LINKED ISSUE #477: Severe performance regression with complex nested function calls
File https://github.com/neovim/nvim-lspconfig/blob/master/scripts/docgen.lua takes 15s to format using latest master, and 133s (> 2min) on #476 branch!

With v13.1 it takes ~14s, and with v12 it takes 7s.

This is a large performance impact. This file should not be taking that long

JohnnyMorganz:
Fixes #477


JohnnyMorganz:
Performance is still not 100% where we want it to be, but this is a big improvement

Issue #465, opened 2022-06-04
Break up long if expressions into multiple lines


evaera:
Before:
```lua
for _, storage in ipairs(self._changedStorage[metatable]) do
	storage[id] = if storage[id] then table.freeze({ old = storage[id].old, new = new }) else record
end
```
After (suggested):
```lua
for _, storage in ipairs(self._changedStorage[metatable]) do
	storage[id] = 
		if storage[id] 
		then table.freeze({ old = storage[id].old, new = new }) 
		else record
end
```

JohnnyMorganz:
This formatting was added in v0.12.0 (was discussed in https://github.com/JohnnyMorganz/StyLua/issues/289)

It will only apply if your line is above the column width though, which I assume in this example it isn't

evaera:
I find the first example really hard to read. I ended up rewriting it to use a regular if statement instead. I think it would have been fine to stay as an ifexpr if it was broken up though.

JohnnyMorganz:
Yeah I agree with you that it's definitely nicer to read when separated on multiple lines.

Not entirely sure what kind of heuristic to apply here though, because it's *technically* correct in that it respects column width, but we want it to trigger earlier. An arbitrary reduction in column width (e.g. column width - 20) is also odd. Relying on existing formatting is also something I want to avoid.

The simplest way right now would be to change your own column width to something like 100 instead of 120, but if you're following a standard then that isn't ideal.

matthargett:
What I'm finding in other large Luau projects is that 120 characters column width as the standard code style just doesn't work with the increased density and flexibility being added to the language. This lines up with a pattern I saw in JS projects migrating TypeScript, and in C++ projects that migrated to C++14 and started to adopt lambdas and other density-per-line increases. In those cases, we also dropped the column width from 120 to 100, or 100 to 80, for the same reasons. (Modern syntax-coloring and grammar-aware code editors were in use in all those projects as well.)

matthargett:
FYI, we merged a change to the Roblox Lua Style Guide today that now recommends using a 100 character line length, versus 120 characters.

JohnnyMorganz:
Coming back to this, I wonder if we should accept the formatter irreversability (like we do for tables), and keep an if expression expanded if it was already expanded in the input code.

I don't think there is any nice heuristic to make this look reasonable, so relying on input is probably our best bet.

Drawback is that it will also suffer from https://github.com/JohnnyMorganz/StyLua/issues/264

Issue #276, opened 2021-10-07
If an argument in a function call is multi-lined, break up all the arguments


archiif:
If an argument within a function call is going to be broken up into multiple lines, the entire series of arguments should also be broken up. So instead of this:

```lua
function_call(arg_1, more_args, function()
    more_function_call(arg)
end)

function_call({
        key = value,
        key = value,
        key = value,
        key = value,
        key = value,
}, arg_1, more_args)
```

I think it's more readable like this:

```lua
function_call(
    arg_1,
    more_args,
    function()
        more_function_call(arg)
    end
)

function_call(
    {
        key = value,
        key = value,
        key = value,
        key = value,
        key = value,
    },
    arg_1,
    more_args
)
```
Again, the styling is referenced from [black](https://github.com/psf/black).
Sorry for the combo of issues, just got a bit excited knowing that there's at least one good Lua formatter out there :)

PR #479, opened 2022-06-26
Collapse simple statements onto single line, behind option

### LINKED ISSUE #161: Format single conditional + LastStmt (`return/break`) "guard" statements on a single line?

### LINKED ISSUE #248: Allow single line function definitions
I think it would be useful to allow definitions of "simple" functions to be placed on single line. So that, for example, the code `func(function() return true end, 2)` will not be formatted as

```lua
func(function()
  return true
end, 2)
```

The notion of "simple function" might be aligned with "simple `if`" from #232. This is also somewhat related to #188.

JohnnyMorganz:
Prototype implementation. Closes #248 and Closes #161 (supersedes #232)
Not confirmed whether this will be added or not

To test, install stylua using
```
cargo install --git https://github.com/JohnnyMorganz/StyLua.git --branch singleline
```
then set the `collapse_simple_statement` option to `FunctionOnly` / `ConditionalOnly` / `Always`

TODO:
- More tests
- Changelog


echasnovski:
This looks great! Thanks!

Here are some cherry-picked "interesting" results of applying this on quite large code base (my [mini.nvim](https://github.com/echasnovski/mini.nvim) Neovim plugin), divided into subjective categories. Overall, the majority of cases are "boring" (something like `function(x) return x == 1 end`) and work perfectly.

Unexpected detail (for me) that I noticed is that my preferred width of 120 might be a bit too much for one line functions. But the ability to have one-line functions (and `if` statements potentially) is so good (for me), that I might consider switching to 100.

<details><summary>Nice</summary>

```diff
-  local comment_parts = vim.tbl_filter(function(x)
-    return x ~= ''
-  end, vim.split(commentstring, '%s', true))
+  local comment_parts = vim.tbl_filter(function(x) return x ~= '' end, vim.split(commentstring, '%s', true))
```

```diff
-Helpers.expect.match = MiniTest.new_expectation('string matching', function(str, pattern)
-  return str:find(pattern) ~= nil
-end, function(str, pattern)
-  return string.format('Pattern: %s\nObserved string: %s', vim.inspect(pattern), str)
-end)
+Helpers.expect.match = MiniTest.new_expectation(
+  'string matching',
+  function(str, pattern) return str:find(pattern) ~= nil end,
+  function(str, pattern) return string.format('Pattern: %s\nObserved string: %s', vim.inspect(pattern), str) end
+)
```

```diff
-    local _, tag_section = toc_entry.parent:has_descendant(function(x)
-      return type(x) == 'table' and x.type == 'section' and x.info.id == '@tag'
-    end)
+    local _, tag_section = toc_entry.parent:has_descendant(
+      function(x) return type(x) == 'table' and x.type == 'section' and x.info.id == '@tag' end
+    )
```

</details>

<details><summary>Questionable</summary>

Most of these seems to be along the lines of "try first pack into single line and then expand other". This leads to `return` and `end` being on separate lines eventually, but in a "crammed" way.

```diff
-function MiniCompletion.default_process_items(items, base)
-  return H.default_config.lsp_completion.process_items(items, base)
-end
+function MiniCompletion.default_process_items(items, base) return H.default_config.lsp_completion.process_items(
+  items,
+  base
+) end
```

```diff
-  ['both'] = function(body, opts)
-    return {
-      top = body.top - 1,
-      bottom = body.bottom + 1,
-      indent = math.max(H.get_line_indent(body.top - 1, opts), H.get_line_indent(body.bottom + 1, opts)),
-    }
-  end,
+  ['both'] = function(body, opts) return {
+    top = body.top - 1,
+    bottom = body.bottom + 1,
+    indent = math.max(H.get_line_indent(body.top - 1, opts), H.get_line_indent(body.bottom + 1, opts)),
+  } end,
```

Previous actually looks better:

```diff
-function H.is_item(x)
-  return type(x) == 'table'
-    and H.is_fun_or_string(x['action'], false)
-    and type(x['name']) == 'string'
-    and type(x['section']) == 'string'
-end
+function H.is_item(x) return type(x) == 'table'
+  and H.is_fun_or_string(x['action'], false)
+  and type(x['name']) == 'string'
+  and type(x['section']) == 'string' end
```

Both nice and questionable:

```diff
-function H.get_unsaved_listed_buffers()
-  return vim.tbl_filter(function(buf_id)
-    return vim.api.nvim_buf_get_option(buf_id, 'modified') and vim.api.nvim_buf_get_option(buf_id, 'buflisted')
-  end, vim.api.nvim_list_bufs())
-end
+function H.get_unsaved_listed_buffers() return vim.tbl_filter(
+  function(buf_id) return vim.api.nvim_buf_get_option(buf_id, 'modified') and vim.api.nvim_buf_get_option(buf_id, 'buflisted') end,
+  vim.api.nvim_list_bufs()
+) end
```

Not as nice with `function aaa() return ... end` as with `aaa = function() return ... end`, but OK:

```diff
-function MiniTest.is_executing()
-  return H.cache.is_executing == true
-end
+function MiniTest.is_executing() return H.cache.is_executing == true end
```

</details>

<details><summary>Case study</summary>

Splitting at `=`:

```diff
-local get_buf_names = function()
-  return child.lua_get('vim.tbl_map(function(x) return vim.fn.bufname(x) end, vim.api.nvim_list_bufs())')
-end
+local get_buf_names =
+  function() return child.lua_get('vim.tbl_map(function(x) return vim.fn.bufname(x) end, vim.api.nvim_list_bufs())') end
```

Mixing functions with and without `return`:

```diff
-local erroring = function(x)
-  return function()
-    error(x, 0)
-  end
-end
+local erroring = function(x) return function()
+  error(x, 0)
+end end
```

Nested function creation:

```diff
-local get_extmarks_short = function()
-  return vim.tbl_map(function(x)
-    return { x.col, x.text }
-  end, get_extmarks())
-end
+local get_extmarks_short = function() return vim.tbl_map(function(x) return { x.col, x.text } end, get_extmarks()) end
```

</details>

JohnnyMorganz:
> Unexpected detail (for me) that I noticed is that my preferred width of 120 might be a bit too much for one line functions. But the ability to have one-line functions (and if statements potentially) is so good (for me), that I might consider switching to 100.

This has been becoming more a theme to be honest, and people are starting to switch their style guides / configs to 100 (https://github.com/JohnnyMorganz/StyLua/issues/465#issuecomment-1163843276).

I think 100 seems to actually be producing much nicer formatting in a lot of places tbh. Too bad its a very big breaking change, but if I were to do it again I would probably set the default to 100

JohnnyMorganz:
@echasnovski made some updates - how do they look now?

JohnnyMorganz:
Added support for collapsing if conditional guard clauses as well. This means the collapse option has changed to be aligned with https://github.com/JohnnyMorganz/StyLua/issues/248#issuecomment-1004049579

echasnovski:
> @echasnovski made some updates - how do they look now?

Gave it a go. Tried all three  I love it even more! Can't back it up with numbers, but subjectively it gave less diff than first iteration. Tried it again on my plugin with all three versions. Here is [the gist](https://gist.github.com/echasnovski/daf45b865f5b41871a9db9a1bb56ed1a) with full diffs (it would be even more if I remove numerous `--stylua: ignore` before single-line functions I already have). `"Always"` gave a merge of those two, as expected.

Couple of observations:
- In comment with suggestion about option name and values it states `collapse_single_statement`, but the option is actually `collapse_simple_statement` (`single` -> `simple`). Might be a bit confusing for others trying this. Also maybe worth editing first comment also (now states that option name is `collapse_mode`)?
- The `"ConditionalOnly"` and `"Always"` make `if` statements single line but lose all indentation. Otherwise, looks great.
- There are couple of instances when function definition was already spanning several lines and couldn't be fit into one line, but it got changed anyway. My personal opinion is that new way looks significantly nicer, but not sure you intended it to be changed. Here is an example diff:
```diff
-  local resolve_data = H.process_lsp_response(H.info.lsp.result, function(response, client_id)
-    -- Return nested table because this will be a second argument of
-    -- `vim.list_extend()` and the whole inner table is a target value here.
-    return { { edits = response.additionalTextEdits, client_id = client_id } }
-  end)
+  local resolve_data = H.process_lsp_response(
+    H.info.lsp.result,
+    function(response, client_id)
+      -- Return nested table because this will be a second argument of
+      -- `vim.list_extend()` and the whole inner table is a target value here.
+      return { { edits = response.additionalTextEdits, client_id = client_id } }
+    end
+  )
```
- Cases with "nested simple functions" don't look as nice, albeit understandable and expected. But those are rare and can be subject for `--stylua: ignore`. Here is what I mean:
```diff
-local get_match = function(hl_group)
-  return vim.tbl_filter(function(x)
-    return x.group == hl_group
-  end, child.fn.getmatches())
-end
+local get_match =
+  function(hl_group) return vim.tbl_filter(function(x) return x.group == hl_group end, child.fn.getmatches()) end
```

JohnnyMorganz:
Thanks for these diffs, this is incredibly helpful!

> values it states `collapse_single_statement`, but the option is actually `collapse_simple_statement`

I think "simple" makes more sense here than "single". I've changed the comment text

> There are couple of instances when function definition was already spanning several lines and couldn't be fit into one line, but it got changed anyway.

Yeah, this was unfortunately not intentional. I agree that the expanded version looks nicer, but there's some cases where we need to keep it collapsed right now #276. Although, since this is behind an option I might keep it, but the unintended deviation may come back to bite me.

> Cases with "nested simple functions" don't look as nice

Agreed, I don't think there are any cases where nested functions *would* look nice collapsed? So it would be likely that a lot of people would be ignoring it. If so, I think its simple enough to prevent it from formatting on single line here

JohnnyMorganz:
Question (this is moreso related to #342):
In this diff, is the preference to keep function expanded, or to collapse it but hang at the equals token:
```diff
-vim.lsp.diagnostic.get_count = function(buf_id, id)
-  return ({ Error = 4, Warning = 3, Information = 2, Hint = 1 })[id]
-end
+vim.lsp.diagnostic.get_count =
+  function(buf_id, id) return ({ Error = 4, Warning = 3, Information = 2, Hint = 1 })[id] end
```

Other than this, I think all the issues brought up have been fixed

JohnnyMorganz:
> Nested "simple function definitions" seems to collapse only the most inner definition. Nice, but to me feels as unnecessary complication via adding "special cases".

It would actually be a more complex (more invasive change) if we wanted to keep the whole thing expanded, instead of letting the innermost definition stay compact. I might look into it more though.

Other than this, is there anything else which particularly stands out to you? Would you (and anyone else viewing) be happy with the output as-is if this PR were to be merged (hypothetically :P)

echasnovski:
> It would actually be a more complex (more invasive change) if we wanted to keep the whole thing expanded, instead of letting the innermost definition stay compact. I might look into it more though.

Interesting. Never would have thought this.

> Other than this, is there anything else which particularly stands out to you? Would you (and anyone else viewing) be happy with the output as-is if this PR were to be merged (hypothetically :P)

Judging by the latest diffs, merge of this PR will be a strong contender for making my week. _Hypothetically_.

Do you have any thoughts about how you'd like to proceed with this? Do you want to get more people's thoughts about it or just time to think about pros and cons of extra option?

JohnnyMorganz:
>  _Hypothetically_.

🤣 

> Do you have any thoughts about how you'd like to proceed with this? Do you want to get more people's thoughts about it or just time to think about pros and cons of extra option?

To be perfectly honest, I think I'm going to go ahead with the option.
Going to give it a little bit of time in case other people have opinions, but otherwise its scheduled for the next release (hopefully in the next few days)

clason:
Thank you for considering this! From a short test, it seems like it only collapses functions with explicit `return`, so
```lua
function() return 1 end
```
but not other statements, so
```lua
function(arg)
  foo(arg)
end
```
Is that deliberate, or am I doing something wrong?

JohnnyMorganz:
Currently deliberate - should single statements also be considered?

clason:
I'd say so, at least, for two reasons:
1. for me, over 90% of these "simple functions" are closures, and these are exactly what I'd like to fold up;
2. it feels inconsistent to only fold specific statements -- which should `return foo` be folded but `foo(bar)` not?

Otherwise this restricted scope should be made _very_ explicit in option name (`collapse_guard_statement`?), PR title, and commit message, otherwise people will get quite confused. ("Closes #248" implies that it implements the more general suggestion, not the current restricted one -- hence my question.)

clason:
(And similarly for conditionals.)

clason:
My personal expectation, given the issue description, would have been "anything that 
1. is a single statement

clason:
should simplify the code, too 😇

echasnovski:
> My personal expectation, given the issue description, would have been "anything that
> 
>     1. is a single statement (including comments into the count)
> 
>     2. fits into the line width after folding
> 
> 
> gets folded, no special exceptions".

To add here: when opening a #248, my initial idea of how it would work was exactly like this. But then it seemed to contradict the original view about "simple conditional" (taken from [here](https://github.com/JohnnyMorganz/StyLua/issues/161#issue-894742877)), so I didn't push for this behavior. That said, having not only `LastStmt` collapsed would be great but current solution is already very much a quality of life improvement.

JohnnyMorganz:
> should simplify the code, too

Unfortunately not - it is a more invasive change [mainly because of the way the formatters have been set up, in particular with adding newlines. tech debt :( ]. It was simpler just having to deal with returns only

To lower the additional complexity, I think it makes sense to narrow down "statements" in general to just a function call or an assignment. I don't think any other statement makes sense, as the rest will always go multiline anyways.

I'll look into adding this still, but it may be another week or so before I get the time to do so

JohnnyMorganz:
Added support for collapsing bodies with simple statements (assignments + function calls).

Let me know how it looks, in particular I would appreciate any interesting test cases

echasnovski:
Looks great! Added [diffs](https://gist.github.com/echasnovski/daf45b865f5b41871a9db9a1bb56ed1a#file-collapse_simple_statement__more_simple-diff) with new changes (first applied ef045e9 commit with option value "Always", staged all, then applied d66f7a4).

Everything seems to work as expected (apart from some "Maybe unintended").

New subjective observations:

- It resulted into higher additional impact than previous approach with only terminal statements had for current `master`.
- Having short `if ... then ... end` on single line right besides short `for ... do ... end` on separate lines felt somewhat inconsistent. Just thinking out loud here, not suggesting additional changes :)
- Having multiple assignments inside `if` block is not very readable. Like `if is_good then a, b = 1, 2 end`. Syntax highlighting certainly helps, so this is only a slight inconvenience.

Looking at diffs reiterated me with these beliefs:
- Using 120 width is too much. Probably, 100 should be good.
- Using `function xxx() ... end` is not as readable as `xxx = function() ... end` when put on single line.

Even with observations and reiterated beliefs, this still feels better for me than previous version with only `LastStmnt` or without them at all.

Some hand picked cases:

<details><summary>Maybe unintended</summary>

Modified `if` block but didn't put on single line:

```diff
   if left == nil then
-    left, right = H.infer_match(
-      { left = left_prev, right = right_prev },
-      { left = left_next, right = right_next },
-      offset
-    )
+    left, right =
+      H.infer_match({ left = left_prev, right = right_prev }, { left = left_next, right = right_next }, offset)
   end
```

Modified expression in already multi line function definition (it is a first expression after comment, so maybe that caused it):

```diff
-  local surroundings = vim.tbl_deep_extend(
-    'force',
-    H.builtin_surroundings,
-    MiniSurround.config.custom_surroundings or {}
-  )
+  local surroundings =
+    vim.tbl_deep_extend('force', H.builtin_surroundings, MiniSurround.config.custom_surroundings or {})
```

</details>

<details><summary>Interesting cases</summary>

```diff
-  local has_signature = b:has_descendant(
-    function(x) return type(x) == 'table' and x.type == 'section' and x.info.id == '@signature' end
-  )
+  local has_signature =
+    b:has_descendant(function(x) return type(x) == 'table' and x.type == 'section' and x.info.id == '@signature' end)
```

```diff
-    if child == nil then
-      child, index = index, #self + 1
-    end
+    if child == nil then child, index = index, #self + 1 end
```

```diff
-T['stat_summary()']['works'] = function()
-  eq(stat_summary(10, 4, 3, 2, 1), { minimum = 1, mean = 4, median = 3, maximum = 10, n = 5, sd = math.sqrt(50 / 4) })
-end
+T['stat_summary()']['works'] =
+  function()
+    eq(stat_summary(10, 4, 3, 2, 1), { minimum = 1, mean = 4, median = 3, maximum = 10, n = 5, sd = math.sqrt(50 / 4) })
+  end
```

```diff
-  expect.error(function()
-    child.lua([[MiniSessions.read('session-absent')]])
-  end, '%(mini%.sessions%) "session%-absent" is not a name for detected session')
+  expect.error(
+    function() child.lua([[MiniSessions.read('session-absent')]]) end,
+    '%(mini%.sessions%) "session%-absent" is not a name for detected session'
+  )
```

</details>

echasnovski:
> I'll have a look at the other examples you gave and see what I can do

To be clear, I don't see anything wrong with them (besides, maybe, the first one had slightly better "before" formatting). They were subjective answer to "...in particular I would appreciate any interesting test cases".

JohnnyMorganz:
I've made a small change to prevent inlining on multiple assignments + fix the eq(stat_summary(... case.
Other than that, I have left it as is.

If there is anything in particular which stands out, let me know - otherwise I will merge this in

echasnovski:
I've checked 00d5bf4 against d66f7a4 and indeed only those two types of changes were reverted. Nothing else.

To me everything looks great! Again :+1:

will:
Really looking forward to this, can't wait to try it out. Thank you!!

qubidt:
FYI, this hasn't been reflected in the README

PR #495, opened 2022-07-10
Fix ignore behavior for --stdin-filepath

MunifTanjim:
## Current Behavior

When the `.styluaignore` file contains:

```ignore
**/ignored_*
```

And `stylua` is run like this:

```sh
cat /home/user/lua-project/ignored_file.lua | stylua --search-parent-directories --stdin-filepath /home/user/lua-project/ignored_file.lua -
```

StyLua still formats the code.

## Expected Behavior

When `--stdin-filepath` is provided, it should respect the `.styluaignore` rules even if the code is read from stdin.

For reference, [Prettier](https://prettier.io) respects the ignore file when [`--stdin-filepath`](https://prettier.io/docs/en/options.html#file-path) is provided.

## Usecase

Editor plugins depend on this behavior. For example, [`null-ls.vim`](https://github.com/jose-elias-alvarez/null-ls.nvim) is broken for `stylua` because of this bug... even if a file is ignored in `.styluaignore`, the file is unexpectedly formatted on save.


src/cli/main.rs
```diff
@@ -341,12 +356,28 @@ fn format(opt: opt::Opt) -> Result<i32> {
...
+                        Some(filepath) => {
+                            let ignore = get_ignore(filepath.parent().unwrap()).unwrap();
+                            match ignore.matched(&filepath, false) {
+                                ignore::Match::Ignore(_) => true,
+                                _ => false,
+                            }
+                        }
+                        None => false,
+                    };
+
```
JohnnyMorganz, line 369: Instead of computing this and still calling `format_string`, could we potentially just exit early here?
```suggestion
                    match &opt.stdin_filepath {
                        Some(filepath) => {
                            let ignore = get_ignore(filepath.parent().unwrap()).unwrap();
                            match ignore.matched(&filepath, false) {
                                ignore::Match::Ignore(_) => continue,
                                _ => (),
                            }
                        }
                        None => (),
                    };
```

And then the changes to `format_string` are no longer necessary as well

JohnnyMorganz: If its changed to this, you could probably also rewrite it as an `if let` statement

src/cli/main.rs
```diff
@@ -341,12 +356,28 @@ fn format(opt: opt::Opt) -> Result<i32> {
                     let tx = tx.clone();
                     let opt = opt.clone();
 
+                    let should_skip_format = match &opt.stdin_filepath {
+                        Some(filepath) => {
+                            let ignore = get_ignore(filepath.parent().unwrap()).unwrap();
```
JohnnyMorganz, line 361: Is it possible for this to ever fail when unwrapping here?
If you know that it shouldn't ever fail unless something is very wrong, then could you use `.expect()` with some sort of error message?

That should only happen if the error here is fatal though, and we want to panic and exit.
If its possible for this to sometimes fail, we should instead handle the error appropriately to allow stylua to continue working with the rest of the files.

Same comment on the other calls to `.unwrap()`

src/cli/main.rs
```diff
@@ -198,6 +204,15 @@ fn format_string(
     }
 }
 
+fn get_ignore(directory: &Path) -> Result<Gitignore> {
+    let file_path = find_ignore_file_path(directory.to_path_buf(), true).unwrap();
+    if let Some(file_path) = file_path {
+        Ok(Gitignore::new(file_path).0)
```
JohnnyMorganz, line 210: The second return to Gitignore::new is typically an error. Instead of ignoring we should surface it up to the user appropriately

src/cli/config.rs
```diff
@@ -48,6 +48,18 @@ fn find_config_file(mut directory: PathBuf, recursive: bool) -> Result<Option<Co
     }
 }
 
+pub fn find_ignore_file_path(mut directory: PathBuf, recursive: bool) -> Result<Option<PathBuf>> {
```
JohnnyMorganz, line 51: Doesn't look like this function can ever error, so it can probably just return `Option<PathBuf>` and remove the errors

src/cli/main.rs
```diff
@@ -198,6 +204,15 @@ fn format_string(
     }
 }
 
+fn get_ignore(directory: &Path) -> Result<Gitignore> {
+    let file_path = find_ignore_file_path(directory.to_path_buf(), true).unwrap();
```
JohnnyMorganz, line 208: Should we always search recursively here? Or only when `--search-parent-directories` is specified?

JohnnyMorganz (changes requested):
Thank you for this! And sorry for the delay in reviewing, I seem to have missed the notification about your PR.

src/cli/main.rs
MunifTanjim: What is the standard pattern for surfacing? Log it to stderr? Panic?

src/cli/main.rs
MunifTanjim: > Instead of computing this and still calling format_string, could we potentially just exit early here?

Format string is doing some other things, e.g. diffing, wrapping with `FormatResult::*`, converting to bytes etc.

Is it okay to bypass all that?

src/cli/main.rs
MunifTanjim: Addressed.

src/cli/config.rs
MunifTanjim: Addressed.

src/cli/main.rs
MunifTanjim: Yep, these calls shouldn't really fail. Addressed.

src/cli/main.rs
JohnnyMorganz: We use anyhow for our errors, which allows us to propagate it upwards and report it properly.

I think gitignore returns a `Vec<Error>` which might need slightly tweaking. You can then `return Err(...)` with the useful info

Check out some other errors, we typically stick `.context` or `.with_context` on the error to provide better info contextual info.

This might be useful: https://docs.rs/anyhow/latest/anyhow/

src/cli/main.rs
JohnnyMorganz: Ah, good point. Not sure what is better to do here, so it's fine to leave it for now, thanks

src/cli/main.rs
JohnnyMorganz: Nevermind, it's an Option<Error>, which is slightly nicer.

You should just be able to check if error exists and then `return Err(error)`. Potentially with some extra context

MunifTanjim: Another question. The errors return from `Gitignore::new` are actually partial errors. Say there's one invalid line in the ignore file, that error will be returned as the second parameter. Should we abort the whole operation because of that partial error? Or just keep going?

JohnnyMorganz: Personally I think it's better to just abort completely, because even if there is a partial error that error could be important for the particular file we are formatting

MunifTanjim: Updated. It will now abort the process with `("failed to parse ignore file: {:#}", err`) if error is returned from `Gitignore::new`.

src/cli/main.rs
```diff
@@ -341,12 +361,31 @@ fn format(opt: opt::Opt) -> Result<i32> {
                     let tx = tx.clone();
                     let opt = opt.clone();
 
+                    let should_skip_format = match &opt.stdin_filepath {
+                        Some(filepath) => {
+                            let ignore = get_ignore(
+                                filepath.parent().expect("cannot get parent directory"),
+                                opt.search_parent_directories,
+                            );
```
JohnnyMorganz, line 369: I haven't tested this, but could `.context("failed to parse ignore file")?` work here?

MunifTanjim: Yep, just tested. That works too. Updated.

src/cli/main.rs
```diff
@@ -341,12 +364,35 @@ fn format(opt: opt::Opt) -> Result<i32> {
                     let tx = tx.clone();
                     let opt = opt.clone();
 
+                    let should_skip_format = match &opt.stdin_filepath {
+                        Some(filepath) => {
+                            let ignore = get_ignore(
+                                filepath.parent().expect("cannot get parent directory"),
+                                opt.search_parent_directories,
+                            )
+                            .context("failed to parse ignore file");
```
JohnnyMorganz, line 373: Sorry, the `?` was meant to be part of it
```suggestion
                            .context("failed to parse ignore file")?;
```

And then you won't need to match on the result

Not sure if this works though

MunifTanjim: Yep, it works and simplifies things 🎉

JohnnyMorganz (approved):
Thanks!

PR #501, opened 2022-07-20
Fix clippy warnings

JohnnyMorganz:


matthargett (approved):
needs rustfmt, then looks good:
```
 --> src/formatters/functions.rs:878:13
    |
878 |             (),
    |             ^^ expected enum `std::option::Option`, found `()`
    |
    = note:   expected enum `std::option::Option<full_moon::ast::types::TypeSpecifier>`
            found unit type `()`
```

PR #512, opened 2022-07-26
Don't hang `local x = function` assignments

### LINKED ISSUE #508: incorrect line break on assignment of function
```lua
exports.ScalarLeafsRule = function(context)
        return {
                Field = function(_self, node)
                        local type_ = context:getType()
                        local selectionSet = node.selectionSet
                        if type_ then
                                if isLeafType(getNamedType(type_)) then
                                        if selectionSet then
                                                local fieldName = node.name.value
                                                local typeStr = inspect(type_)
                                                context:reportError(
                                                        GraphQLError.new(
                                                                ('Field "%s" must not have a selection since type "%s" has no subfields.'):format(
                                                                        fieldName,
                                                                        typeStr
                                                                ),
                                                                selectionSet
                                                        )
                                                )
                                        end
                                elseif not selectionSet then
                                        local fieldName = node.name.value
                                        local typeStr = inspect(type_)
                                        context:reportError(
                                                GraphQLError.new(
                                                        ('Field "%s" of type "%s" must have a selection of subfields. Did you mean "%s { ... }"?'):format(
                                                                fieldName,
                                                                typeStr,
                                                                fieldName
                                                        ),
                                                        node
                                                )
                                        )
                                end
                        end
                end,
        }
end
```

image

JohnnyMorganz:
Fixes #508

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed var expression with trailing comments on initial prefix being collapsed leading to malformed formatting ([#509](https://github.com/JohnnyMorganz/StyLua/issues/509))
+- Assignments of functions using `local x = function()` syntax will no longer ever hang unnecessarily at the equals token. ([#508](https://github.com/JohnnyMorganz/StyLua/issues/508))
 
 ## [0.14.1] - 2022-07-21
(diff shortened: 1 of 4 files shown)


JohnnyMorganz:
Not the right fix - luau regression test shows we will make something worse...

JohnnyMorganz:
Will revisit later. This isn't correct fix and the problem is looking a lot more involved.

PR #549, opened 2022-08-21
Don't expand call with nested comment

### LINKED ISSUE #543: Don't expand function call containing multiline comment unnecessarily
```lua
call(item, --[[param=]] false)
```
... gets formatted as ...
```lua
call(
	item, --[[param=]]
	false
)
```
but this is not necessary

### LINKED ISSUE #561: Inline parameter comments may be wrapped to wrong line
For example, this:
```lua
self:doTheThing(what, --[[rightAway = ]] true)
```

Is rewritten by Stylua to:
```lua
self:doTheThing(
    what, --[[rightAway = ]]
    true
)
```

It would be preferable to distinguish these short inline parameter comments and not wrap at all.

JohnnyMorganz:
Fixes #543 
Fixes #561


JohnnyMorganz:
Potential negative impact from this change... Need to rethink

JohnnyMorganz:
Closing this out as the change produces negative side effects. The original formatting is reasonable to remain, compared to the new formatting

matthargett (approved):

PR #558, opened 2022-08-31
Update pyproject to support installing on M1 machines

IamTheFij:
The format is a bit different for macOS arm64 from the others in that
it's the only one specifying an arch. As a result, any x86_64 archs are
mapped to an empty string and the arm64 one has a prefixed `-` to
separate it from the system.


JohnnyMorganz:
Thanks!

PR #559, opened 2022-08-31
Rename release targets and add linux-aarch64

IamTheFij:
I'm unsure if this will work in the build workflow.


JohnnyMorganz:
Looks reasonable to me, but I will double check over the weekend

JohnnyMorganz (commented):
Can you also modify https://github.com/JohnnyMorganz/StyLua/blob/main/stylua-npm-bin/binary.js to point to the aarch64 version for linux?

I think we also need to update the VSCode extension (https://github.com/JohnnyMorganz/StyLua/blob/main/stylua-vscode/src/util.ts#L18), but the vscode extension also needs fixing for macos aarch64

IamTheFij:
I could do that, yea. Overall, it would probably be easier to template these out if the formatting of the names was consistent. Eg. everything being `stylua-{system}-{arch}.zip` rather than having to optionally append the arch with a prefixed hyphen.

What do you think about that? Then I'd update this PR to overhaul the naming and update all downloaders (python and node) to use the new format.

JohnnyMorganz:
> I could do that, yea. Overall, it would probably be easier to template these out if the formatting of the names was consistent. Eg. everything being `stylua-{system}-{arch}.zip` rather than having to optionally append the arch with a prefixed hyphen.
> 
> What do you think about that? Then I'd update this PR to overhaul the naming and update all downloaders (python and node) to use the new format.

Yeah in hindsight this is probably what should've happened. The only thing is there are a lot of consumers and changing how the binary was named already caused headache the first time when we removed the version

ViViDboarder:
Yea, is it better to just change now and get it over with though? Could probably have some period of time where the release maintains both files as well to make the transition easier.

JohnnyMorganz:
Yeah ok, let's do it. We can append "-x86_64" to the end, and keep artifacts with the old name temporarily

.github/workflows/release.yml
```diff
@@ -89,6 +95,18 @@ jobs:
           asset_name: ${{ matrix.artifact-name }}.zip
           asset_content_type: application/zip
 
+      # TODO: Remove this after deprecation notice
+      - name: Upload Binary to Release aliases
+        if: ${{ matrix.artifact-alias == "" }}
```
JohnnyMorganz, line 100: I think this condition should be negated?
```suggestion
        if: ${{ matrix.artifact-alias != "" }}
```

JohnnyMorganz (commented):
Could you also update the changelog so I don't forget about this when I create the next release?

IamTheFij:
Ok. Fixed the condition, updated the pyproject to use the new formatting, and fixed the merge conflict by merging in master.

Looks like the lint for VSCode extension failed. I'm installing npm and prettier to fix it, but I'm also making a new branch to add pre-commit hooks to the repo to make it easier to catch things like this before pushing. So far, I'm just adding the same checks that already exist in GitHub Actions. That will be a different PR.

JohnnyMorganz (approved):
Thanks!

JohnnyMorganz:
Fixed the syntax issue but looks like compiling linux aarch64 fails: https://github.com/JohnnyMorganz/StyLua/runs/8307187443?check_suite_focus=true

IamTheFij:
Addressed in #572

PR #566, opened 2022-09-04
Improve comments within function calls

### LINKED ISSUE #565: regression after merging inline comment in argument list
after merging #549 

```lua
if x(
  --[[ we don't use it ]]
   true,
  90210
  --[[
      color wheel is time-reversed
  ]],
  --[[ frobnikate the widget ]]
  false,
  true
  --[[ spin the tesla coils ]]
) then
  print("hello")
end
```

image

JohnnyMorganz:
Building on top of #549 

Fixes #565

### DIFF
--- a/tests/snapshots/tests__standard@function-call-multiline-comment.lua.snap
+++ b/tests/snapshots/tests__standard@function-call-multiline-comment.lua.snap
@@ -7,3 +7,25 @@ expression: format(&contents)
 
 call(item, --[[param=]] false)
 
+call(--[[ we don't use it ]] true)
+
+call(
+	--[[
+		this comment spans
+		multiple lines
+	]]
+	false
+)
+
+x(
+	true,
+	90210,
+	--[[
+		color wheel is time-reversed
+	]]
+	--[[ frobnikate the widget ]]
+	false,
+	true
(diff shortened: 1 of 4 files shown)


tests/snapshots/tests__standard@function-call-multiline-comment.lua.snap
```diff
@@ -7,3 +7,25 @@ expression: format(&contents)
...
+	]]
+	false
+)
+
+x(
+	true,
+	90210,
+	--[[
+		color wheel is time-reversed
+	]]
```
matthargett, line 25: Interesting that it moves these outside of the comma, but maybe not a big deal?

matthargett (approved):

tests/snapshots/tests__standard@function-call-multiline-comment.lua.snap
JohnnyMorganz: Tbh, I think the likelihood of a comment in this particular position (after expression but before the comma) is very low, so not sure it matters too much.

I moved it afterwards for simplicity, but I made an assumption here - let me see if it still works in its current position

PR #568, opened 2022-09-05
Add search parent dirs config for VSCode extension

### LINKED ISSUE #560: VSCode extension setting for config path(s)
When using a multi-root workspace we currently have to duplicate the same `stylua.toml` into each workspace root, it would be great to have a setting to point the VSCode extension to a config at the root of the repository instead.

A respective config for `.styluaignore` may also be necessary but I'm not sure if the CLI accepts a path arg for that?

filiptibell:
Closes #560


JohnnyMorganz (approved):
Thanks!
I really should get a new extension release out soon...

PR #575, opened 2022-09-13
Add `--output-format=summary`

### LINKED ISSUE #573: [Feature Request] `--output-format=summary`
Hey there!

Especially in CI but also locally, I would love to have an overview of which fails failed instead of always getting the full diff as the output, so some sort of summary that could be glanced over would be great.

JohnnyMorganz:
Closes #573 

image
image


Stanzilla:
Oh perfect, thank you!

Issue #407, opened 2022-03-12
Error parsing valid Lua with label at end of function


Feez:
I was experimenting with using StyLua on minified Lua, and found that StyLua fails to parse this valid Lua when minified.

source:
```lua
local function test()
    goto end_thing

    local x = 1

    ::end_thing::
end
```

-> minified:
```lua
local function a()goto b;local c=1::b::end
```

with error:
```bash
error: could not format file minified.lua: error parsing: unexpected token `end`. (starting from line 1, character 40 
and ending on line 1, character 43)
additional information: expected identifier after `::`
```

If I add a semicolon after `local c = 1` in the minified Lua then it parses as expected.

JohnnyMorganz:
I assume this is through downloading the binaries from GitHub releases? This is unfortunately because the binary is currently built with both Lua 5.2 and Luau parsing enabled, which causes this conflict. This was actually reported before too: https://github.com/JohnnyMorganz/StyLua/issues/309#issuecomment-983614651

The only option right now is to install the binary using cargo `cargo install stylua --features lua52`

Feez:
Sure, I'll have to do some setup of cargo.

I believe the output of this should probably be a feature request in being able to enable/disable these features at runtime for the ones that are installed, with the addition of having this error message (and others if they exist) to ask the user to provide the parser feature they intend to use.

JohnnyMorganz:
I definitely agree with having a runtime flag to do this, that is the ideal scenario.

Unfortunately though the issue is upstream in our parser dependency, so it needs to be solved there

Feez:
Maybe in the meantime, could provide additional binaries in the GitHub releases with the respective features compiled into them individually, as well as the default one for all. There seems to be quite a lot that goes into installing Cargo on a system like Windows and I can imagine some developers would have already stopped trying at this point.

JohnnyMorganz:
I also agree on your second point, having Github release binaries for all possible flags would be good (#103). But as I mentioned in https://github.com/JohnnyMorganz/StyLua/issues/309#issuecomment-983614651, right now there is potentially an issue with toolchain managers where they don't know which binary to install.

I might test it out one release though, and publish all the binaries and see if anything breaks

JohnnyMorganz:
The next release of StyLua will give the ability to configure the syntax at runtime in your `.stylua.toml` file.

```toml
syntax = "Lua52"
```

Or alternatively specify it on the command line, `stylua --syntax lua52 file.lua`

[...]

PR #592, opened 2022-09-26
Add `serialize`, `fromstr` and `wasm-bindgen` features

RubixDev:
While working on creating a [dprint](https://dprint.dev/) plugin for integrating StyLua ([dprint-plugin-stylua](https://github.com/RubixDev/dprint-plugin-stylua)), I found few things were missing in StyLua to make the integration easier:

1. A `serde::Serialize` impl for the config types
2. A `FromStr` impl for the config enum types (using `strum::EnumString`)
3. Disabling the `wasm-bindgen` crate for correct compilation

I turned these three things into features and left the defaults just like they were before.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
           sudo apt install ${{ matrix.linker }}
 
       - name: Build Binary (All features)
-        run: cargo build --verbose --locked --release --features lua53,lua53,lua54,luau --target ${{ matrix.cargo-target }}
+        run: cargo build --verbose --locked --release --features lua52,lua53,lua54,luau --target ${{ matrix.cargo-target }}
         env:
           CARGO_TARGET_DIR: output
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
           sudo apt install ${{ matrix.linker }}
 
       - name: Build Binary (All features)
-        run: cargo build --verbose --locked --release --all-features --target ${{ matrix.cargo-target }}
+        run: cargo build --verbose --locked --release --features lua53,lua53,lua54,luau --target ${{ matrix.cargo-target }}
         env:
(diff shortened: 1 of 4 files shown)


JohnnyMorganz:
We currently run `--all-features` in our release CI job. We probably need to change that to explicitly only select `--features lua52,lua53,lua54,luau`

RubixDev:
`wasm-bindgen` is part of the default features, as it only has any effect for wasm32 targets anyways. I will change the workflow files to specifically list the `luaXX` features.

Also would you recommend enabling all the `luaXX` features for the dprint plugin wasm binary? As far as I understand it adds support for those Lua versions while still supporting 5.1?

JohnnyMorganz:
> Also would you recommend enabling all the `luaXX` features for the dprint plugin wasm binary? As far as I understand it adds support for those Lua versions while still supporting 5.1?

For simplicitly, this is what we do for our release assets, and its also what is done in the packaging manifests for stuff like homebrew.

For the most part, this is fine. But note that there can be conflicts between the different Lua versions. The main things I know of is:
- `goto` can be used as an identifier in 5.1/Luau, but not in 5.2+. All features enabled doesn't respect this
- Label syntax `::label::` conflicts with Luau's type assetions `expr :: type` - https://github.com/JohnnyMorganz/StyLua/issues/407

Ideally we would select the syntax we are using through some sort of configuration at runtime, but currently the features are behind compile time flags so we can't do this.

If possible, my recommendation would be to produce wasm binaries for each different syntax version, then using config select the correct one to use. This comes with disadvantages though (you now need to package 5x the wasm binary). Having all features enabled should typically be fine for everyday use.

RubixDev:
Ok thanks, I think I'll just enable all Lua features then and maybe make it configurable in the future.

JohnnyMorganz (approved):
Sorry for the delay, looks good to me, thanks!

I wonder if we even need separate features for `serialize` and `fromstr`, since they are small enough. For now I'll just leave it as-is though

Issue #506, opened 2022-07-25
Potentially unwanted behavior for function calls assignment


MunifTanjim:
## Description

I'm not sure if this is intentional. If it's intentional, it sure looks odd. Behavior of `v0.13.x` was much nicer.

### Input

```lua
  local height = math.max(math.min(#items, defaults(options.max_height, vim.o.lines - 7)), defaults(options.min_height, 1))
```

### Output (v0.13.x)

```lua
  local height = math.max(
    math.min(#items, defaults(options.max_height, vim.o.lines - 7)),
    defaults(options.min_height, 1)
  )
```

### Output (v0.14.x)

```lua
  local height =
    math.max(math.min(#items, defaults(options.max_height, vim.o.lines - 7)), defaults(options.min_height, 1))
```

JohnnyMorganz:
This was intentional: https://github.com/JohnnyMorganz/StyLua/pull/342#issuecomment-1166613067

But what you report seems to be an unfortunate side effect. A lot of the other cases now look nicer, but I also think it looks better here when it is expanded.

I wonder what kind of heuristic we can apply here

MunifTanjim:
<img width="1475" alt="image" src="https://user-images.githubusercontent.com/8050659/180828778-21f44929-c089-4b3b-988b-f31f6e79f156.png">

<img width="1475" alt="image" src="https://user-images.githubusercontent.com/8050659/180828842-3d84b410-45b8-4e2e-85ba-14471df0e719.png">

<img width="1475" alt="image" src="https://user-images.githubusercontent.com/8050659/180828879-ef4e9a88-78ef-4cfa-8138-03610bba6676.png">

<img width="1475" alt="image" src="https://user-images.githubusercontent.com/8050659/180828908-2d539d63-847a-4aea-a0de-1b6136ef80cc.png">

<img width="1459" alt="image" src="https://user-images.githubusercontent.com/8050659/180829136-30b9e15f-7fbc-4323-86c5-da98f765b027.png">

<img width="1459" alt="image" src="https://user-images.githubusercontent.com/8050659/180829187-cf721e76-633b-4525-9124-433fc4f4803c.png">


Prettier's behavior seems pretty reasonable.

JohnnyMorganz:
I think the a potential solution to solve this is to implement some sort of "complexity" heuristic, to decide whether we should hang at equals or not.
For example, the expression mentioned in OP is quite hard to read, so should be multilined.

[...]

PR #600, opened 2022-10-10
Dont hang `local x = function` assignments (Take 2)

### LINKED ISSUE #595: stop hanging function assignment at `=`
0.14 introduced a feature where it will hang at the `=` instead of the argument list for many kinds of constructs. There's a particular scenario where this negatively impacts readability:
```lua
exports.createResource = function(
        glitch: (Input) -> Thenable<Value>,
        hasInput: (Input) -> Key,
        config: Config?
): Pleasing<Input, Key, Value>
        config = config or {}
        local pleasing
        pleasing = {
                clear = function(): ()
                        entries[pleasing] = nil
                end,
        }
        return pleasing
end
```

currently gets transformed into this, which in the integration case triggers a cascade of indents
image

JohnnyMorganz:
Extension on #512 

Closes #595


matthargett:
could we extend this to function call assignments as well?
```lua
     local furryAndFidelity9 = getFurriesAndFidelitiesNames(
             context,
             creamedFurriesAndFidelitiesNames,
             parentType2,
             selectionSet2
     )
```

MunifTanjim:
@matthargett this is the related issue: https://github.com/JohnnyMorganz/StyLua/issues/506

I'm still using `v0.13.x` because of this.

JohnnyMorganz:
PR has been on hold because I need to decide whether to accept the new diff on neovim code in comparison test

matthargett:
FWIW, the Neovim diff looks great to me! ;)

Issue #620, opened 2022-11-15
Remove "opinionated" and allow more config?


JohnnyMorganz:
A discussion which continues to happen offline is the decision against config options for StyLua, so I've decided to open an issue on it.

My main goal has always been to have "sane" defaults in the formatting, but it is impossible to suit every case.

I wonder if being an "opinionated" formatted is still beneficial. It's main selling point has been to reduce bikeshedding, but in practice, I'm not sure if that actually rings true.

[Rustfmt](https://github.com/rust-lang/rustfmt) highlights an example where it's very configurable, but works pretty well (and from what I've seen, most people don't actually bother configuring, but maybe because there are so many options 😅).

Maybe we should just allow more (reasonable) configurations?

Betelgeuse1:
I just used stylua for the first time in a project and the available configuration didn't strike me.
I think being opinionated is beneficial, at least for me. Keeps me from thinking too much about small code appearance details.

What type of extra configuration you had in mind ?

JohnnyMorganz:
> I think being opinionated is beneficial, at least for me. Keeps me from thinking too much about small code appearance details.

Agreed IMO.

I am still unsure about what we do need to configure, some examples include https://github.com/JohnnyMorganz/StyLua/issues/555 or https://github.com/JohnnyMorganz/StyLua/issues/276.
There are also very minor ones such as https://github.com/JohnnyMorganz/StyLua/issues/313, which I don't think are completely worthwhile to change the philosophy for, but I respect that people have different style ideals

Betelgeuse1:
I don't know if it is possible in current code implementation (pretty sure it's not), but maybe stylua should stay opinionated with a way to implement custom rules as [go-ruleguard](https://github.com/quasilyte/go-ruleguard).

That way, a user who really want a certain rule can always write it themselves.

LastTalon:
I think rustfmt benefits heavily from pushing a [conventional](https://en.wikipedia.org/wiki/Convention_over_configuration) approach to writing code using it. Its generally very well accepted that the defaults are very good and sensible defaults and the agreement within the ecosystem is an important aspect of using the formatter.

I think the same can be used here as well. Formatting Lua code is unfortunately still very young and it'll take time for people to catch on and it'll take time for people to realize that certain patterns are very useful for writability and not very useful for readability and a formatter allows you to always favor readability.

Kampfkarren:
[...]

PR #628, opened 2022-12-04
Add `brace_spacing` option

### LINKED ISSUE #313: [Request option] Don't add space within curly braces?
So I have an example here: 
```
example = {vaue1, value 2}
```
When formatted: 
```
example = { value1, value2 }
```
However, I prefer this: 
```
example = {value1, value1}
```
Can you consider make an option to do this?

bbc2:
Braces, used for table construction, are typically used with spaces inside them:

```lua
t = { "content-0" }
```

There can be another style, however, which consists in sticking the braces to the content:

```lua
t = {"content-0"}
```

This work adds a configuration parameter `brace_spacing: bool` to control the spacing inside table constructors and enable the use of the second style. This is similar to [Prettier's bracket spacing option](https://prettier.io/docs/en/options.html#bracket-spacing).

Which style is better is debatable, of course, and so is the configurability. In my quick research, I listed what the formatters I know do:

- rustfmt (Rust): space
- OCamlFormat (OCaml): space, configurable
- Prettier (JavaScript, TypeScript): space, configurable
- Black (Python): no space
- Clang-Format (C, C++): no space, configurable

Closes #313


bbc2:
To be honest, I have mixed feelings about adding options to formatters (maintainability, fragmentation of the community). Even if this is rejected, I hope it will shed some light on https://github.com/JohnnyMorganz/StyLua/issues/620.

Regardless, I will definitely use this tool so thank you for it! Also, I found it very easy to modify.

JohnnyMorganz:
An interesting thing that prettier does, is that it doesn't have spaces around array braces, but does in object braces (to be fair, maybe because they are 2 different syntax constructs in JS/TS, but not in Lua).

I wonder if that would be an interesting middle ground?

Regardless, for all options I tend to not make them boolean but instead enums, so that they can be extended later.

No promises on whether this will be merged or not, but just some thoughts!

bbc2:
> An interesting thing that prettier does, is that it doesn't have spaces around array braces, but does in object braces (to be fair, maybe because they are 2 different syntax constructs in JS/TS, but not in Lua).
>
> I wonder if that would be an interesting middle ground?

I'm not sure because in Lua you can mix the two forms in the same literal:

```lua
table = {"x", "y", a = "z"}
-- or
table = { "x", "y", a = "z" }
```

Lua is pretty wild. :)

> Regardless, for all options I tend to not make them boolean but instead enums, so that they can be extended later (and also makes it slightly easier to understand what the option does)

Done! I also added the `--brace-spacing` CLI option for the sake of completeness. Easy enough.

michaeldougal:
Just came here to say that I've been pretty happy with StyLua so far except for this one missing feature. Please consider merging!

I'll note too that while Roblox (and I realize this is not just about Roblox, but they do constitute a significant group of Lua(u) developers) doesn't take an official stand on it in their style guide, most documentation examples follow the form of having no outer edge spacing, so that is likely to be the form that many developers on the platform learn and use.

As for fragmenting the community or causing divides by adding options, I wouldn't worry about that. StyLua is presented as an opinionated code formatter, and I don't see why we can't use it to implement our own opinionated formatting. Customization options seem to only be beneficial to me. We are also already given options for different parts like quote formatting, so this is entirely appropriate imo.

Tartopohm:
> Just came here to say that I've been pretty happy with StyLua so far except for this one missing feature. Please consider merging!
Agreed!

> As for fragmenting the community or causing divides by adding options, I wouldn't worry about that.
Agreed as well. If anything, allowing for more configuration should have the opposite effect of fragmentation, by allowing people with different opinions to use the same tool, as long as the configuration is agreed upon at the project level.

raiguard:
I will voice my support for this, it's the one thing in stylua that I don't like.

premek:
just a note: editorconfig proposes:
`spaces_around_brackets  (none, inside, outside, both)`

It's mentioned in the section: 
[Ideas for Domain-Specific Properties](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#ideas-for-domain-specific-properties)
The following properties are not intended to be implemented by EditorConfig.
This is simply a brainstorm of domain-specific properties that could be supported by some tools that rely on EditorConfig files.

PR #631, opened 2022-12-18
Collapse `goto` as a simple statement

### LINKED ISSUE #618: `collapse_simple_statement` for goto
using `collapse_simple_statement = "Always"`:

```lua
	-- won't collapse
	if key == "s" then
		goto continue
	end
	-- on the other hand something like this does collapse
	local modes = { "n", "v" }
	if key == "p" then modes = { "n" } end
	-- ...
```

JohnnyMorganz:
Closes #618

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


JohnnyMorganz:
> Integration/snapshot test?

Test was added inline since it needs to configure settings https://github.com/JohnnyMorganz/StyLua/pull/631/files#diff-2ad10836ccce5ac2056d5679cc92449d9ff9094d4ff5c5803f65b5dd1d52ef19

PR #645, opened 2023-02-04
Add default `editorconfig` feature

### LINKED ISSUE #75: [Suggestion] Support .editorconfig
Given that [.editorconfig file](https://editorconfig.org/) provides some settings that are supported by stylua, it could be handy to read that file to avoid duplicating these settings.

LEI:
This adds support for [EditorConfig](https://spec.editorconfig.org/) via [ec4rs](https://docs.rs/ec4rs/latest/ec4rs/) including [domain-specific properties](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties) and should close #75 once the special cases are settled:

- `end_of_line` = `line_endings`
  cr = Unix (Mac)
- `indent_size`/`tab_width` = `indent_width`
- `indent_style` = `indent_type`
- `max_line_length` = `column_width`
- `quote_type` = `quote_style`
  double = AutoPreferDouble
  single = AutoPreferSingle
  auto = StyLua default
- `call_parentheses`
- `collapse_simple_statement`
- `sort_requires`

EditorConfig properties `charset`, `insert_final_newline` and `trim_trailing_whitespace` are not supported.

The `.editorconfig` file in the current directory is loaded if no `.stylua.toml` file was found: not sure if it's the desired behavior and how it should handle different directories. I used `pub mod editorconfig` in `src/lib.rs` in order to run tests on a real directory structure.


src/editorconfig.rs
```diff
@@ -0,0 +1,316 @@
...
+
+#[cfg(feature = "editorconfig")]
+pub fn load_editorconfig(config: Config, directory: &Path) -> Config {
+    // HACK: used to detect filetype, this is not a real file path or glob
+    let path = directory.join("*.lua");
+
+    // Read the EditorConfig files that would apply to a file at the given path
+    let mut cfg = ec4rs::properties_of(path).unwrap();
+
+    cfg.use_fallbacks();
```
LEI, line 16: Not sure about this:
https://docs.rs/ec4rs/latest/ec4rs/struct.Properties.html#method.use_fallbacks
https://docs.rs/ec4rs/latest/src/ec4rs/fallback.rs.html

README.md
```diff
@@ -193,7 +193,8 @@ StyLua is **opinionated**, so only a few options are provided.
 ### Finding the configuration
 
 The CLI looks for `stylua.toml` or `.stylua.toml` in the directory where the tool was executed.
-If not found, the default configuration is used.
+If not found, the default configuration is used and it looks for `.editorconfig` in the current directory.
```
JohnnyMorganz, line 196:
```suggestion
If not found, we search for an `.editorconfig` file, otherwise fall back to the default configuration.
```

tests/test_editorconfig.rs
```diff
@@ -0,0 +1,53 @@
...
+#[cfg(feature = "editorconfig")]
+fn test_local() {
+    insta::glob!("inputs-editorconfig/local/*.lua", |path| {
+        let contents = std::fs::read_to_string(path).unwrap();
+        insta::assert_snapshot!(format(&contents, "local"));
+    })
+}
+
+// Subdirectory without an EditorConfig file
+#[ignore]
```
JohnnyMorganz, line 45: Is there a reason this test is just ignored?

src/cli/config.rs
```diff
@@ -143,7 +146,12 @@ pub fn load_config(opt: &Opt) -> Result<Config> {
 
                     // Fallback to a default configuration
                     debug!("config: falling back to default config");
-                    Ok(Config::default())
+                    let config = Config::default();
+
+                    #[cfg(feature = "editorconfig")]
+                    load_editorconfig(config, &current_dir);
```
JohnnyMorganz, line 152: Is the output of this actually being used, since `config` isn't mutable?

src/editorconfig.rs
```diff
@@ -0,0 +1,294 @@
...
+    std::path::Path,
+};
+
+#[cfg(feature = "editorconfig")]
+pub fn load_editorconfig(config: Config, directory: &Path) -> Config {
+    // HACK: used to detect filetype, this is not a real file path or glob
+    let path = directory.join("*.lua");
+
+    // Read the EditorConfig files that would apply to a file at the given path
+    let mut cfg = properties_of(path).unwrap();
```
JohnnyMorganz, line 19: We should gracefully handle the errors here rather than unwrapping

src/editorconfig.rs
```diff
@@ -0,0 +1,294 @@
...
+fn parse_integer(raw: &RawValue) -> Result<usize, std::num::ParseIntError> {
+    raw.filter_unset().into_str().parse::<usize>()
+}
+
+#[cfg(feature = "editorconfig")]
+fn load_editorconfig_overrides(config: Config, cfg: &Properties) -> Config {
+    let mut new_config = config;
+
+    // EditorConfig end_of_line
+    match parse_string(cfg.get_raw::<EndOfLine>()).as_str() {
```
JohnnyMorganz, line 39: Is there a reason why in all of these that we use the raw version rather than the preprocessed properties and their corresponding enums https://docs.rs/ec4rs/latest/ec4rs/property/index.html

tests/inputs-editorconfig/.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+
+# FIXME:
```
JohnnyMorganz, line 7: What is this FIXME for?

JohnnyMorganz (commented):
Thanks for submitting this!

A few things:
- What happens when there doesn't exist any `.editorconfig` file at all? It looks like we "apply" the default properties from `ec4rs`, but are they the same as the stylua defaults? Is it possible to detect that no `.editorconfig` file was found, and just use `Config::default()`? Maybe there should be a test here to ensure the final config == Config::default()


- Maybe we should search in parent directories for `.editorconfig` as well when `--search-parent-directories` is enabled
- Looks like there are some unrelated changes, probably to make clippy happy. I guess its fine to leave them in this PR

tests/test_editorconfig.rs
JohnnyMorganz: Nevermind, just saw the reason at the end of your original PR message.

LEI:
- The properties are now parsed only if they are found in the .editorconfig file.
  There may a better way to check if each key is matching, maybe with some kind of hash.
  I couldn't find how to use `::key()` in `match` cases (not static). 

- I think the core library [looks up parent directories](https://docs.rs/ec4rs/latest/src/ec4rs/file.rs.html#94) itself (until `root = true`), as it is expected with EditorConfig.

The ignored test in `tests/test_editorconfig.rs` for patterns like `[directory/file.lua]` was failing because `find_editorconfig` needs the exact file name.

**Each file** in a same directory could have a different configuration, though it is probably not a common use case.

Not sure how (multiple?) files provided in command-line arguments should be handled.

src/cli/config.rs
LEI: It was used by `use_fallbacks`, which is now removed.

src/editorconfig.rs
LEI: I thought `cfg.get` would not lowercase the values (but it does), and this is now replaced with `cfg.iter` + `value.parse`, which does too.

LEI:
EditorConfig overrides are applied in `src/cli/config.rs` only if there is a standard input (`-` in args), based on `stdin-filepath` or the current directory.

Otherwise, the configuration override is loaded for each file in `src/cli/main.rs`, maybe the first check could be done around [here](https://github.com/JohnnyMorganz/StyLua/blob/main/src/cli/main.rs#L408) for consistency.

JohnnyMorganz:
> EditorConfig overrides are applied in src/cli/config.rs only if there is a standard input (- in args), based on stdin-filepath or the current directory.

Does this mean if I run `stylua path/to/foo.lua` it will still look for an editorconfig file? If not, is there a reason this functionality is removed?

JohnnyMorganz:
> Otherwise, the configuration overrides are loaded for each file in src/cli/main.rs, maybe the stdin check could be done around [here](https://github.com/JohnnyMorganz/StyLua/blob/main/src/cli/main.rs#L408) for consistency.

Yeah, I think this would be better, and then `src/cli/config.rs` can be left alone.

> The properties are now parsed only if they are found by the library.
Maybe there is a cleaner way to check if each key is matching, with some kind of hash.
I couldn't find how to use ::key() in match cases (not static).

Is it necessary to loop through the properties? It might just be simpler if its something like

```rs
fn load(config: Config, properties: &Properties) -> Config {
    let mut new_config = config;
    
    // Handle standard keys
    if let Ok(end_of_line) = properties.get::<EndOfLine>() {
         // ...
    }
    if let Ok(indent_size) = properties.get::<IndentSize>() {
         // ...
    }
    // ...

   // Handle custom keys
```

Also, for the custom keys, is it worth creating a custom enum for them and impl PropertyKey + PropertyValue? If its not too necessary, feel free to leave it

> Should `no_call_parentheses` and `tab_width` be parsed?

Sorry - I missed this. I don't think `no_call_parentheses` should be included (its deprecated). `tab_width` should be I think though, especially if `indent_size` points to UseTabWidth

I think we should also have a test with both an `.editorconfig` file and a `stylua.toml` file to ensure that latter is used.

Could you also update the CHANGELOG (add something into the Added section).

Sorry for the long review time, thank you once again for adding this!

LEI:
Fixes:

- Moved the logic to `main`, `config::load_config` now returns `None` if no `stylua.toml` was found.
- Changed the properties matching to a `if` sequence and extracted the `property_choice` macro for enums.
- `tab_width` should now be correctly handled (`properties.use_fallbacks()` is commented).

TODO:
- [ ] Improve `stylua.toml` + `.editorconfig` test
- [ ] Test CRLF (attempt in `.gitattributes`)
  > warning: in the working copy of `tests/snapshots/test_editorconfig__stylua_toml@eol-crlf.lua.snap`, LF will be replaced by CRLF the next time Git touches it
- [ ] Add support for `sort_requires`?

src/editorconfig.rs
```diff
@@ -0,0 +1,374 @@
...
+use ec4rs::{
+    properties_of,
+    property::{EndOfLine, IndentSize, IndentStyle, MaxLineLen, TabWidth, UnknownValueError},
+    rawvalue::RawValue,
+    Error, Properties, PropertyKey, PropertyValue,
+};
+use std::path::Path;
+
+// Read the EditorConfig files that would apply to a file at the given path
+pub fn parse(config: Option<Config>, path: &Path) -> Result<Option<Config>, Error> {
```
JohnnyMorganz, line 11: Is it necessary for the input and output here to be `Option<Config>`? It looks like `None` is only input in the stdin version, but can't we just pass the base `config` the same as the filepath version

src/cli/main.rs
```diff
@@ -447,17 +470,26 @@ fn format(opt: opt::Opt) -> Result<i32> {
...
-                                    },
-                                ),
-                            )
+                            #[cfg(not(feature = "editorconfig"))]
+                            let c = config;
+                            #[cfg(feature = "editorconfig")]
+                            let c = is_default_config
+                                .then(|| {
+                                    editorconfig::parse(Some(config), &path)
+                                        .expect("could not parse editorconfig")
```
JohnnyMorganz, line 479: Rather than `.expect`'ing here, we should propogate it back up (through the `tx.send`) so its handled nicely using `anyhow`, and also add in some context. This is mainly because it also appropriately JSON-ifies the error message.

Something along the lines of the following could work (untested):
```rs
#[cfg(feature = "editorconfig")]
let used_config = match is_default_config {
    true => editorconfig::parse(Some(config), &path).context("could not parse editorconfig"),
    false => Ok(Some(config)),
};

tx.send(
    used_config
        .and_then(|used_config| {
            format_file(
                &path,
                used_config.unwrap_or(config),
                range,
                &opt,
                verify_output,
            )
        })
        .map_err(|error| {
            ErrorFileWrapper {
                file: path.display().to_string(),
                error,
            }
            .into()
        }),
)
.unwrap()
```

You may need to convert `parse` to return an `anyhow::Result` instead (like `format_file` does)

src/editorconfig.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    config
+}
+
+// Extracted from ec4rs::property
+macro_rules! property_choice {
+    ($prop_id:ident, $name:literal; $(($variant:ident, $string:literal)),+) => {
+        #[derive(Clone, Copy, PartialEq, Eq, Debug)]
+        #[repr(u8)]
+        #[doc = concat!("The [`",$name,"`](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#",$name,") property.")]
```
JohnnyMorganz, line 103:
```suggestion
```

src/lib.rs
```diff
@@ -7,6 +7,8 @@ use wasm_bindgen::prelude::*;
 
 #[macro_use]
 mod context;
+#[cfg(feature = "editorconfig")]
+pub mod editorconfig;
```
JohnnyMorganz, line 11: Full public visibility to editorconfig is probably not ideal, since it shouldn't be an API consumers rely on. I don't think there is much we can do about it though, `pub(crate)` is probably too restrictive.

JohnnyMorganz (commented):
Thanks! Just a few more comments which are hopefully a bit more minor.
Also, could we add a CLI option `--no-editorconfig` to ignore editorconfig completely like prettier does: https://prettier.io/docs/en/cli.html#--no-editorconfig

> - Add support for sort_requires?

Yeah, I guess we can do

src/editorconfig.rs
LEI: It was used for logging if no properties were found, this is removed now.

src/cli/main.rs
```diff
@@ -403,12 +407,31 @@ fn format(opt: opt::Opt) -> Result<i32> {
...
+                        #[cfg(not(feature = "editorconfig"))]
+                        let c = config;
+                        #[cfg(feature = "editorconfig")]
+                        let c = path
+                            .and_then(|path| {
+                                editorconfig::parse(None, &path)
+                                    .expect("could not parse editorconfig")
+                            })
+                            .unwrap_or(config);
+
```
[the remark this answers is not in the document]
LEI: I was not able to use `cwd` after `move` (`use occurs due to use in closure`), maybe there is an obvious fix.

src/cli/main.rs
LEI: Done for the second check, unsure about [this line](https://github.com/JohnnyMorganz/StyLua/blob/262dbce6d749fa9e0e2980b4e3a05aef45ce8a3f/src/cli/main.rs#L429) for the first check (stdin), should `map_err` be used twice around here?

src/editorconfig.rs
LEI: [Also removed](https://github.com/JohnnyMorganz/StyLua/pull/645/commits/550f40b7e31472e44ef787dcb2cc54fafe85ed02#diff-bde938c835cc94f24ee5b19bd4ac7eecf0c80eae46ffe2df066bdd720d4b11b1L104) `#[allow(missing_docs)]` or does it have any use?

LEI:
Reordered the functions in `editorconfig.rs` and added `sort_requires = true`, or should it be `sort_requires_enabled`?

src/editorconfig.rs
JohnnyMorganz: Don't think so, we don't deny missing docs at the crate level. Fine to remove

src/cli/main.rs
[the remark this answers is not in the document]
JohnnyMorganz: Ah fair. I'm not sure from first glance without playing around to try and please the compiler.

Do we actually even need the cwd here? Could the glob `*.lua` be sufficient? Or might that miss some config

JohnnyMorganz:
> Reordered the functions in `editorconfig.rs` and added `sort_requires = true`, or should it be `sort_requires_enabled`?

Probably best to use `sort_requires_enabled` for the sake of forwards compatibility in case future options are added to the sort requires code mod

src/cli/main.rs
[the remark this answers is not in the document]
LEI: Good point: `cwd` was not needed, problem solved.

Note that `*.lua` is **not** a glob, `a.lua` could be used instead: it would also match `[*.lua]` (the just core library expects a file path).

If `[*.luau]` or another extension is defined in `.editorconfig`, it won't be taken into account when formatting from stdin without `--stdin-filepath`.

LEI:
Future options could work like this and it saves a few characters, unless I am missing something:
```
sort_requires = true
sort_requires_new_option = true
```

JohnnyMorganz:
> Future options could work like this and it saves a few characters, unless I am missing something:
> 
> ```
> 
> sort_requires = true
> 
> sort_requires_new_option = true
> 
> ```

Yeah you're right. Let's leave it as-is then

src/cli/main.rs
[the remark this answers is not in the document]
JohnnyMorganz: Fair enough. I think its reasonable to expect the user to provide stdin filepath in most cases (typically stdin is used for editor integrations or scripts)

JohnnyMorganz (approved):
Thanks for tackling this longstanding issue!

I'm happy with the general implementation here. I don't know if you have anymore outstanding TODOs. If not, I'll probably merge by the end of this week and get a release out then

src/lib.rs
LEI: Indeed, there are [ways](https://users.rust-lang.org/t/pub-only-for-tests/28153) around it but it seems more involved.

I just pushed some cleanup, hopefully fixing the case where the input file would be left modified if `test_stylua_toml` failed, and removing the EOL test attempt (too tricky).

My last concern would be `MaxLineLen::Off => config.with_column_width(0)` since I did not test if `0 === off`.

LEI: `max_line_length = off` was not working, fixed in the latest commit using `usize::MAX` instead of `0`.

LEI:
The test for max line length is now updated and commits are rebased.

Should be ready to go if you are okay with `pub mod editorconfig`.

JohnnyMorganz (approved):
Looks good to me, thank you, and sorry for the delay!

mindbound:
What is the reason why `insert_final_newline ` is not supported?

JohnnyMorganz:
Stylua does not respect the option itself - it will always add a newline.

Although, it should parse the option and ignore it - an error is not useful if you are using the editor config for multiple languages

mindbound:
That's extremely odd - in both my neovim setup and running from command line, `stylua` does the exact opposite: it always _removes_ the final newline from the formatted file. That's why I was asking the previous question - i.e., hoping that there is an option to prevent this behaviour.

JohnnyMorganz:
Huh, that's definitely weird. I just tried stylua from the command line right now and it added a new line
image

mindbound:
I'm getting the exact opposite: after adding a newline after `local x = 1`, it gets removed on formatting. I'm using Arch, `stylua` installed from the official `extra` repository, version 0.19.1.

JohnnyMorganz:
Sorry, I would have no clue how to diagnose further as I do not have a setup such as yours. All our [test cases](https://github.com/JohnnyMorganz/StyLua/tree/main/tests/snapshots) have a final newline so I would assume it is something to do with your setup.

Stupid question - but could it be whatever you are using to render your file is hiding the final newline? Note that stylua does trim multiple final newlines into only one newline, that may be hidden by your editor.

mindbound:
That was my initial thought as well but apparently not, this is really what happens:

image

JohnnyMorganz:
That is really weird. Sorry I don't have a better answer for you, I'm curious what is causing this

JohnnyMorganz:
Wait a second - isn't that output correct? Your initial test.lua file has two newlines, which we trim down to 1

`wc -l test.lua` would report 0 if there was no trailing newline, and you would see something like this the first line in cat:

image

PR #653, opened 2023-02-11
Sort requires

### LINKED ISSUE #364: Sort requires

JohnnyMorganz:
Take 2 on the sort requires system (supersedes #454)
We take a much simpler approach here.

The following assumptions are made when using this codemod:
- All requires are pure and have no side effects: resorting the requires is not an issue
- Only requires at the top level block are to be sorted
- Requires are of the form `local NAME = require(REQUIRE)`, with only a single require per local assignment

Requires sorting works in the following way:
- We group consecutive requires into a "block". If we hit a line which is a non-require or empty, we close the old block and start a new one.
- Requires are sorted only within their block. This allows us to solve the assumption of depending on local variables (if there is a local variable in between requires, it would split them into two separate blocks, so a require will always be after any local variable it uses)
- Blocks remain in-place in the file.
- If any of the block contains an ignore comment (`-- stylua: ignore`), or falls outside of the specified range, then the whole block will be skipped

We also sort services of the form `local NAME = game:GetService(EXPR)` in the same way (treated as a separate group)

Closes #364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added support for "sort requires", which sorts top-level statements of the form `local NAME = require(EXPR)` lexicographically on `NAME`.
+  We do this by treating a group of consecutive requires as a "block", and then sorting **only within** the block. Any other statement, or an empty line, between require statements will split the group into two separate blocks (and can be used to separate the sorting). A block of requires will not move around the file.
+  Roblox Luau statements of the form `local NAME = game:GetService(EXPR)` will also be sorted separately.
+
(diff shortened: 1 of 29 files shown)


Kampfkarren:
How is capitalization formatted? I have a lot of code that mixes both capital and lowercase initials

JohnnyMorganz:
It is case sensitive. It will sort all uppercase variables before lowercase ones - I've added a test to show this

Kampfkarren:
☹️ i wish it didn't, but i've never been able to get a consistency i like anyway

JohnnyMorganz:
> ☹️ i wish it didn't, but i've never been able to get a consistency i like anyway

We could reasonably add an "ignore case" option here if it better suits

PR #655, opened 2023-02-24
Add Dockerfile

eberkund:
It would be nice if this tool could also be distributed as a Docker image in addition to the existing packages.

This would make it much easier to build multi-platform images without manually checking the architecture and downloading the right binary.


JohnnyMorganz (commented):
Sounds useful, thank you!

Maybe this should also have a quick mention in the README about installation and usage?

eberkund:
Hi, thanks for the quick response. Your project has been very helpful for me. 

I opened this PR as a starting point, I'd be happy to update the README as well. Ideally the image can also be published to Dockerhub via GitHub Actions if you're open to it. Do you have any preferred approach for this?

JohnnyMorganz:
No preference. I've never actually used Docker, so not sure. If we need to setup a docker hub account, I can do that once the basic push action is ready. I know github also has a container registry we could push to. Let me know what you need

.github/workflows/docker.yml
```diff
@@ -0,0 +1,43 @@
+name: Docker
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
JohnnyMorganz, line 7: Do we need to run this action on every push? Could it be part of the `release.yml` workflow instead?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,43 @@
...
+          DOCKER_METADATA_PR_HEAD_SHA: "true"
+        with:
+          images: JohnnyMorganz/StyLua
+          tags: |
+            type=semver,pattern={{version}}
+
+      - uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN
```
JohnnyMorganz, line 32:
```suggestion
          password: ${{ secrets.DOCKERHUB_TOKEN }}
```

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM rust:1.67 AS build
+ARG CARGO_FLAGS="--all-features --release"
+
+WORKDIR /app
+COPY . ./
+RUN cargo build ${CARGO_FLAGS}
+
+FROM gcr.io/distroless/cc
+COPY --from=build /app/target/release/stylua /
+CMD ["./stylua"]
```
JohnnyMorganz, line 10: I don't know too much about Docker, but is it worth configuring an `ENTRYPOINT` as well? It seems that allows it to be used through `docker run`

JohnnyMorganz:
Also, lets build to the GitHub Container Registry as well just so its available as much as possible https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages

eberkund:
> Also, lets build to the GitHub Container Registry as well just so its available as much as possible https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages

Hey, I don't think there is any reason to publish to Docker Hub and GHCR and it wouldn't restrict its available having it in one or the other.

Dockerfile
eberkund: You can still use `docker run` with `CMD`, I think this is the right way. From what I understand `ENTRYPOINT` is more for when you still want the `ENTRYPOINT` to run in addition to specifying an additional command which I don't think is the case for a CLI tool.

You can try like this:

```
$ docker build . -t stylua
$ docker run stylua /stylua "*.lua"
```

JohnnyMorganz (approved):
Sorry for the delay, looks reasonable to me!
Not sure if there is any way to test this, but I've added the relevant tokens

eberkund:
> Sorry for the delay, looks reasonable to me! Not sure if there is any way to test this, but I've added the relevant tokens

Well I tested the Dockerfile by building it locally. You added tokens for Docker Hub or GHCR? I should update the PR to push to the right repository and I suppose to test I can just modify the triggers or create a test job?

JohnnyMorganz:
> > Sorry for the delay, looks reasonable to me! Not sure if there is any way to test this, but I've added the relevant tokens
> 
> 
> 
> Well I tested the Dockerfile by building it locally. You added tokens for Docker Hub or GHCR? I should update the PR to push to the right repository and I suppose to test I can just modify the triggers or create a test job?

I did it for docker hub. I think it should be fine to trigger an example yeah

eberkund:
Okay, can you trigger the workflow then? I don't think I have permissions but the workflow should work now.

JohnnyMorganz:
Thank you!

PR #680, opened 2023-04-17
Don't remove parentheses around optional type

### LINKED ISSUE #679: StyLua removes necessary parentheses around type
The parentheses around the type `{ number }?` are removed by StyLua's formatting, which produces a Luau error:
image
The following code was used to reproduce the issue:
```lua
local x = {} :: {} & ({ number }?)
```

JohnnyMorganz:
Fixes #679


JohnnyMorganz:
Need to be more clever in its removal - looks like we are gonna have to get messy with context

PR #735, opened 2023-07-10
vscode-stylua: search configuration file for Lua files not in workspace

taquangtrung:
Hi,

The current VSCode StyLua extension is unable to search for configuration files in parent directories of a Lua file not in the current workspace.

I fixed that issue by setting the `cwd` parameter to the parent directory of the currently opened file.

Can you consider merging this PR?

Thanks!

### DIFF
--- a/stylua-vscode/src/extension.ts
+++ b/stylua-vscode/src/extension.ts
@@ -1,4 +1,5 @@
 import * as vscode from "vscode";
+import * as path from "path";
 import { formatCode, checkIgnored } from "./stylua";
 import { GitHub } from "./github";
 import { StyluaDownloader } from "./download";
@@ -84,7 +85,8 @@ export async function activate(context: vscode.ExtensionContext) {
         const currentWorkspace = vscode.workspace.getWorkspaceFolder(
           document.uri
         );
-        const cwd = currentWorkspace?.uri?.fsPath;
+        // const cwd = currentWorkspace?.uri?.fsPath;
+        const cwd = path.dirname(document.uri.fsPath);
 
         if (await checkIgnored(document.uri, currentWorkspace?.uri)) {
           return [];


LastTalon:
Can you explain the use-case of this and what it's solving? This seems like it's probably better solved by changing the StyLua binary to search for configuration recursively upward rather than adding an additional one-level-up check in the extension.

stylua-vscode/src/extension.ts
```diff
@@ -84,7 +85,8 @@ export async function activate(context: vscode.ExtensionContext) {
         const currentWorkspace = vscode.workspace.getWorkspaceFolder(
           document.uri
         );
-        const cwd = currentWorkspace?.uri?.fsPath;
+        // const cwd = currentWorkspace?.uri?.fsPath;
```
JohnnyMorganz, line 88:
```suggestion
```
No need to keep the old line

JohnnyMorganz (commented):
Looks good, thanks for the fix!

JohnnyMorganz (changes requested):
Actually, wait a second, I might be wrong here.

This does fix one case, but then breaks another case.

Right now the extension works by using the root workspace folder as the cwd.
That works for the following (typical) case:

```
foo/
    bar.lua
baz.lua
stylua.toml
```

Your PR I assume is meant to fix the following case:
```
foo/
    bar.lua
    stylua.toml
baz.lua
```
Where we are formatting bar.lua using the nested stylua.toml

This, however, will break the original typical case, since Stylua by default does not search parent directories

taquangtrung:
Hi, thanks for reviewing the PR!

It won't break the original case if `stylua.searchParentDirectories` is set to `true`.

The typical use case here is that suppose  you're working in the workspace `~/workspace-1` but need to edit a Lua file in a folder outside it, say `~/workspace-2/foo/bar.lua` and the configuration file is `~/workspace-2/.stylua.toml`

The current code that use `~/workkspace-1` as CWD won't be able to format `~/workspace-2/foo/bar.lua`

To resolve your example, I think the better solution is to set ` stylua.searchParentDirectories` to `true`, in addition to the new changes in this PR.

JohnnyMorganz:
> To resolve your example, I think the better solution is to set  stylua.searchParentDirectories to true, in addition to the new changes in this PR.

This has unintended side-effects, which is why the setting is not enabled by default in the first place.

I can see the issue the PR is trying to solve. What if instead we check if the file we are trying to format falls outside of the current workspace, then apply this change? If yes, then use its parent directory as the cwd. If no, then use the current workspace as cwd, as it does right now.

taquangtrung:
@JohnnyMorganz: I just updated the PR according to your suggestion. Can you take a look?

JohnnyMorganz (approved):
Looks like it should work, thank you! I haven't tested it myself yet - did it work as needed for you?

Issue #807, opened 2023-10-29
pre-commit setup failure: KeyError: 'git-url'


ferronn-dev:
With the following Dockerfile:

```Dockerfile
FROM mcr.microsoft.com/vscode/devcontainers/python
WORKDIR /repo
COPY .pre-commit-config.yaml .
RUN pip install pre-commit
RUN git init
RUN pre-commit
```

and .pre-commit-config.yaml:

```yaml
repos:
- repo: https://github.com/JohnnyMorganz/StyLua
  rev: v0.18.2
  hooks:
  - id: stylua-github
```

Docker build fails. Here's the pre-commit log:

### version information

```
pre-commit version: 3.5.0
git --version: git version 2.42.0
sys.version:
    3.12.0 (main, Oct  3 2023, 02:01:29) [GCC 12.2.0]
sys.executable: /usr/local/bin/python
os.name: posix
sys.platform: linux
```

### error information

```
An unexpected error has occurred: CalledProcessError: command: ('/root/.cache/pre-commit/repon_d6qjtg/py_env-python3.12/bin/python', '-mpip', 'install', '.')
return code: 1
stdout:
    Processing /root/.cache/pre-commit/repon_d6qjtg
      Installing build dependencies: started
      Installing build dependencies: finished with status 'done'
      Getting requirements to build wheel: started
      Getting requirements to build wheel: finished with status 'done'
      Preparing metadata (pyproject.toml): started
      Preparing metadata (pyproject.toml): finished with status 'error'
stderr:
      error: subprocess-exited-with-error
      
      × Preparing metadata (pyproject.toml) did not run successfully.
      │ exit code: 1
      ╰─> [17 lines of output]
          Prepare meta /tmp/pip-modern-metadata-6b23go_n None
          Traceback (most recent call last):
            File "/root/.cache/pre-commit/repon_d6qjtg/py_env-python3.12/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
              main()
            File "/root/.cache/pre-commit/repon_d6qjtg/py_env-python3.12/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
              json_out['return_val'] = hook(**hook_input['kwargs'])
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #808, opened 2023-10-31
Fix compatibility with release-gitter 2.2.0

rouge8:
Fixes https://github.com/JohnnyMorganz/StyLua/issues/807


JohnnyMorganz (approved):
Thanks for taking a look and fixing this!

I don't use the precommit hook myself so I haven't tested this - I'm assuming it works given that you've submitted the patch :)

PR #818, opened 2023-11-01
Fix SortRequireConfig on Wasm build

Barocena:
This PR adds Constructor function for SortRequireConfig and exposes to Wasm build


JohnnyMorganz (approved):
Thanks!

PR #826, opened 2023-11-19
Add status bar instead of error notifications (VSC extension)

DervexHero:
This change allows users to hide formatter errors in the extension settings. These error notifications are especially annoying when using `Format On Save` or `Format On Paste` options so it makes sense to disable them optionally.

image


JohnnyMorganz:
Yeah, I don't particularly like these errors too much. I was wondering whether we should just remove them completely and either replace them with:

a) proper VSCode diagnostics. But is it even worth it? Seems more like a role for a language server / linter than the formatter

b) an icon on the status bar - similar to how Prettier does it

DervexHero:
I think the second option is better. I'm going to implement this in like half an hour.

DervexHero:
Okay so I added status bar item and removed unnecessary setting, take a look:

Success:
image

Failure:
image

JohnnyMorganz:
Looks pretty great, thanks for taking the time!

I'll try it myself next week and hopefully get this merged in.

(And wow, that error message is so verbose..., I never really realised)

LastTalon:
> a) proper VSCode diagnostics. But is it even worth it? Seems more like a role for a language server / linter than the formatter
> 
> b) an icon on the status bar - similar to how Prettier does it

People typically use prettier with eslint in my experience, so they get diagnostics through that. It would be best if selene could handle a stylua plugin.

JohnnyMorganz:
Thank you very much for contributing it. So sorry for the delay in testing it out, took a bit more than a week 😅 

One minor thing I noticed was when you switch files on StyLua error, it still shows up as an error - it should probably clear itself when the active text editor changes. Maybe the item should even hide on non-Lua files.

I really like it displaying in the Status Bar. I did find something new in VSCode though, a "Language Status Item": https://code.visualstudio.com/api/references/vscode-api#LanguageStatusItem, which seems like another great spot to put this, as I don't think we need to take up extra space. For example, for eslint:
image

And this also gives the user the option to pin it to the bar if they want.

I took the liberty to build off your work and do the above changes, hope you don't mind. Thank you once again!

PR #834, opened 2023-12-13
Add new release artifact x86_64-unknown-linux-musl

andros21:
Hi, thanks for `stylua` formatter! I use it a lot within neovim.

The aim of this PR is to support distro like Alpine Linux (on amd64 arch),
building and attaching a new artifact for the target `x86_64-unknown-linux-musl` 
when releasing with `release.yml` workflow

From local tests, the build process for this new target finished without errors ... only one warning
```log
 warning: dropping unsupported crate type `cdylib` for target `x86_64-unknown-linux-musl`
```
related to the `[lib]` part of `Cargo.toml`

Let me know if I need to do something ... maybe trigger some CI to test `stylua` works also for this new target

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,6 +36,9 @@ jobs:
             artifact-alias: stylua-linux
             cargo-target: x86_64-unknown-linux-gnu
           - os: ubuntu-20.04
+            artifact-name: stylua-linux-x86_64-musl
+            cargo-target: x86_64-unknown-linux-musl
+          - os: ubuntu-20.04
             artifact-name: stylua-linux-aarch64
             cargo-target: aarch64-unknown-linux-gnu
             linker: gcc-aarch64-linux-gnu


JohnnyMorganz:
Sorry for the delay, and thanks for the contribution!

The only thing I am worried about here is the format of the name. Unfortunately we didn't leave much flexibility in the way the artifact name is structured, and I know many tools hardcode the structure to determine what asset to download. We do this ourselves in multiple places:

- VSCode extension: https://github.com/JohnnyMorganz/StyLua/blob/5de9a1da8a8dc628b98240e12ffe24c7fdbbea6e/stylua-vscode/src/util.ts#L18C1-L52
- GitHub Action: https://github.com/JohnnyMorganz/stylua-action/blob/b703a90701bbd044c75258647f5aecd8ddb504e1/src/stylua.ts#L41-L76
- Our NPM package: https://github.com/JohnnyMorganz/StyLua/blob/5de9a1da8a8dc628b98240e12ffe24c7fdbbea6e/stylua-npm-bin/binary.js#L13-L39
- The precommit hook installer: https://github.com/JohnnyMorganz/StyLua/blob/5de9a1da8a8dc628b98240e12ffe24c7fdbbea6e/pyproject.toml#L5-L15

And we will need to make sure this new package name doesn't affect those at the minimum, let alone the third party packagers (but it is impossible to control all of them).

But..., the concern here is only that these tools don't accidently pick up the new artifact, and just use the current ones still. So, we just need to structure it in that way. I think maybe `stylua-linux-x86_64-musl` might be better instead to not hit a few of those regexes.

andros21:
Fully agree. I was not aware of this jungle of regexes 😅
Now I change it

What do you think instead to move the alias `artifact-alias: stylua-linux` now linked with **linux-gnu** target to the new **linux-musl** target? It should be more portable on linux platforms cause statically compiled against musl-libc, a minimum version of glibc is not needed

JohnnyMorganz:
I think we should leave the alias for now. The alias artifact is actually deprecated and I intend to remove it at the next major release.

JohnnyMorganz (approved):
Thank you! I don't anticipate any problems, but let's see at the next release. If something comes up, I can just remove the zip from the release artifacts temporarily.

PR #839, opened 2023-12-23
Add option for formatting with spaces between function names and arguments

### LINKED ISSUE #832: Add option to add spaces in function definitions
One style choice I've made extensively across many projects that clashes with `stylua` is to add spaces in function definitions between the function name and arguments.

```lua

-- like this
local function foo (arg) end
foo(6)

-- or like this
local foo = function (arg) end
foo(6)

-- not this
local function(arg) end
foo(6)

-- and not this
local foo = function(arg) end
foo(6)
```

This style gives some visual differentiation between function definitions and function calls. In these simple examples it is relatively obvious, but it is possible for closures to get a bit confusing. Obviously this is an opinionated choice, but I'm not the only one to make this choice. I got it from other projects and just thought it was valuable enough to continue.

I can understand if people don't want this to be the default (as I would set it) but at least an option would be nice.

alerque:
Closes #832

~~So far just some CLI options and tests. Work on implementation pending feedback in issue.~~ All options implemented and tested.


alerque:
I'm using this in production including setting the config option it introduces all over the place. The *results* are what I want to see. If the *implementation* here isn't the best way to do it or needs adjusting to be included I'm happy to adapt, but it would be nice if this could get upstreamed so I don't have to push by fork of StyLua on anybody contributing to my Lua projects ;-)

JohnnyMorganz:
I think I'm going to get some movement on #854 this weekend (which I want to merge before next release) so I also hope to review this too. Sorry for the wait!

alerque:
Sure fair enough. From an implementation standpoint there are probably better ways to do this (I was quite unfamiliar with the code base) even with the current parser, and improved full-moon interfaces may open up even better ways. From the perspective of the end user feature introduction the two important parts are the name of the new config option key and values paired with their respective expected test outputs. Everything in between is implementation details that could be refactored at any time.

Personally I'm only interested in the `SpaceAfterFunction::Definitions` option here, but besides `Always` and `Never` I also went ahead and implemented the `Calls` variant because I've seen code bases that use that. The default is of course `Never` so as not to cause churn in existing users.

src/formatters/functions.rs
```diff
@@ -1158,7 +1165,8 @@ pub fn format_function_declaration(
         shape
     )
     .update_leading_trivia(FormatTriviaType::Append(leading_trivia));
-    let formatted_function_name = format_function_name(ctx, function_declaration.name(), shape);
+    let formatted_function_name = format_function_name(ctx, function_declaration.name(), shape)
+        .update_trailing_trivia(FormatTriviaType::Append(function_definition_trivia));
 
     let shape = shape + (9 + strip_trivia(&formatted_function_name).to_string().len()); // 9 = "function "
```
JohnnyMorganz, line 1171: A point worth mentioning is that our "shape" calculation here (and potentially elsewhere) will not take into account the fact that a space was added after the function name (since we explicitly strip the trivia afterwards). Shape is used to store the amount of line budget used so far and how many characters we have left on a line. This opens up a very small edge case where the definition line now hits `column_width + 1` characters because of the space, but shape only sees `column_width` and therefore we don't reformat correctly. It's quite a minor edge case though, unsure if it is worth rectifying until someone brings it up

README.md
```diff
@@ -259,6 +259,7 @@ StyLua only offers the following options:
...
+| `space_after_fuctions` | `Never`            | Specify whether to add a space beetween the function name and parentheses. Possible options: `Never`, `Definitions`, `Calls`, or `Always`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
JohnnyMorganz, line 262:
```suggestion
| `space_after_functions` | `Never`            | Specify whether to add a space between the function name and parentheses. Possible options: `Never`, `Definitions`, `Calls`, or `Always`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```

README.md
```diff
@@ -271,6 +272,7 @@ indent_width = 4
 quote_style = "AutoPreferDouble"
 call_parentheses = "Always"
 collapse_simple_statement = "Never"
+space_after_fuctions = "Never"
```
JohnnyMorganz, line 275:
```suggestion
space_after_functions = "Never"
```

JohnnyMorganz (commented):
Implementation itself looks good to me, thank you!

>  the name of the new config option key and values paired with their respective expected test outputs

I'm trying to think if the key name is complete descriptive of what is actually happening here for someone with no context. The space isn't really after the function, so maybe it should be `space_in_functions` or `space_in_functions_before_parentheses`. Not sure though, naming is always hard.

alerque:
> Implementation itself looks good to me, thank you!
> 
> > the name of the new config option key and values paired with their respective expected test outputs
> 
> I'm trying to think if the key name is complete descriptive of what is actually happening here for someone with no context. The space isn't really after the function, so maybe it should be `space_in_functions` or `space_in_functions_before_parentheses`. Not sure though, naming is always hard.

Yes naming is hard. I'm happy to refactor this to another name if you think it best, but I'm not convinced there are better options. Actually instead of `space_after_functions` the more correct name might be `space_after_function_names`. It just seems wordy.

I don't like `space_in_functions` or `space_in_function_before_parenthesis` because both suggest a different location entirely in my mind.

src/formatters/functions.rs
alerque: I'd be happy to fix the edge case if I had any idea where/how, but I wasn't actually that clear on the internals of how that was being done or where extra info should be passed to get it back. Given what you say here I'm guessing the same edge case might exist for some other scenarios unrelated to this PR, no?

alerque:
The lint error seems to be unrelated to this PR and more related to a new Rust version. In fact it might even be a false positive for the new lint, I know there was some talk of false positives in new docstring lints. I haven't looked into whether this is one of them, but it seems that should be in a separate PR anywhere.

alerque:
The force-push just now was to add a missed bit to override the configs if CLI arguments were present. The option was working from config (or editor config) and accepted from the CLI but silently ignored.

alerque:
Beep beep, gentle ping. I keep wanting to contribute style overhauls and CI jobs for projects using this, but it's kind of limiting to have it stuck in a PR. Many of the projects I hope to see adopt this *already use this style* (for example Pandoc) and it would be unreasonable churn to expect them to change.

JohnnyMorganz:
I am so so sorry for this delay, I haven't had the chance to work on this project for a while.

I renamed the option to `space_after_function_names`, although it's a bit more wordy, it is a bit more descriptive and I don't think wordiness should matter too much in the config file. Also went ahead and added it to the changelog.

Thank you for taking the time to contribute this, sorry it took almost a year to get it in. I'm going to try get the full-moon update PR ready soon to get this out into a release

PR #870, opened 2024-06-28
Apply clippy suggestions

alerque:
My PR #839 is throwing a lint error, but the culprit exists on the default branch without the PR so in the interest of separation of concerns, I'm filing the fix for that separately.


JohnnyMorganz (approved):
Thank you!

PR #908, opened 2024-10-27
Update wasm-bindgen and wasm-pack

### LINKED ISSUE #888: Update wasm-bindgen`and wasm-pack to resolve ReferenceError
Hello! I'm trying to use the [npm package](https://www.npmjs.com/package/@johnnymorganz/stylua) via [vite-plugin-wasm](https://github.com/Menci/vite-plugin-wasm), but I'm running into the following error:

```
Uncaught ReferenceError: Cannot access 'memory' before initialization
    at stylua_lib_bg.js?v=4e2ee103:671:42
(anonymous) @ stylua_lib_bg.js?v=4e2ee103:671
```

I believe this error is the same as https://github.com/Menci/vite-plugin-wasm/issues/8#issuecomment-1193187290. It sounds like that was resolved by https://github.com/rustwasm/wasm-bindgen/pull/2965, but that fix is only present in wasm-bindgen 0.2.82 and onwards (but this project is using 0.2.81). I was wondering if you'd have any problem with updating things to incorporate the fix.

mtoohey31:
I believe this should resolve the wasm-related errors I mentioned in #888. My specific project appears to work correctly after these changes. Other than building the code and using it my project though, I haven't tested anything else, so lmk if there's anything else I need to check. Thanks!

Closes #888.

### DIFF
--- a/wasm/build-wasm.sh
+++ b/wasm/build-wasm.sh
@@ -1,10 +1,7 @@
 # TODO: Ensure that version is up to date
 cp README.md wasm/
 cp LICENSE.md wasm/
-npx wasm-pack@0.10.3 build --target web --out-dir wasm/stylua.web -- --features lua52,lua53,lua54,luau
-
-# workaround for bundler usage
-echo "export { getImports as __getImports, finalizeInit as __finalizeInit }" >> wasm/stylua.web/stylua_lib.js
+npx wasm-pack@0.13.0 build --target web --out-dir wasm/stylua.web -- --features lua52,lua53,lua54,luau
 
 # bundle for node CommonJS
 npx rollup@4.9.5 wasm/src/stylua_lib_node.cjs --file wasm/stylua_lib.cjs --format cjs
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -650,9 +650,9 @@ dependencies = [
 
 [[package]]
 name = "once_cell"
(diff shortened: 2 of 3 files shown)


wasm/build-wasm.sh
```diff
@@ -1,10 +1,7 @@
 # TODO: Ensure that version is up to date
 cp README.md wasm/
 cp LICENSE.md wasm/
-npx wasm-pack@0.10.3 build --target web --out-dir wasm/stylua.web -- --features lua52,lua53,lua54,luau
-
-# workaround for bundler usage
-echo "export { getImports as __getImports, finalizeInit as __finalizeInit }" >> wasm/stylua.web/stylua_lib.js
```
JohnnyMorganz, line 7: How come we are removing this line?

JohnnyMorganz (commented):
Thanks for this!

wasm/build-wasm.sh
mtoohey31: Good question, now that you mention this I've realized that other files in the parent directory seem to depend on these exports. The issue is the file which is generated by the latest version of wasm-pack doesn't contain `getImports` or `finalizeInit`. I'm not exactly clear on what these were actually for though...

So I've converted this to a draft for now. I'm kinda busy right now so it might be a little bit before I can dig into this and figure out what those were for.

PR #909, opened 2024-11-12
Update documentation of 'opinionated' to reflect project goals

### LINKED ISSUE #620: Remove "opinionated" and allow more config?

alerque:
Closes #620

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 	</h1>
 </div>
 
-An opinionated code formatter for Lua 5.1, 5.2, 5.3, 5.4 and [Luau](https://roblox.github.io/luau/), built using [full-moon](https://github.com/Kampfkarren/full-moon).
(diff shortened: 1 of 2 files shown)


JohnnyMorganz (approved):
Nice, thanks!

PR #917, opened 2024-11-17
Add option to allow formatting files listed in .gitignore

### LINKED ISSUE #895: Add .gitignore ignore cli flag/option

JohnnyMorganz:
Closes #895 

Not sure what the best name for this is, I've gone with `--allow-gitignore` for now

### DIFF
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -304,7 +304,7 @@ fn format(opt: opt::Opt) -> Result<i32> {
     }
 
     walker_builder
-        .standard_filters(true)
+        .standard_filters(!opt.allow_gitignore)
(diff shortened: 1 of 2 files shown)


ValdezFOmar:
> Not sure what the best name for this is, I've gone with `--allow-gitignore` for now

May I suggest `--no-ignore` as an alternative? At least [`fd`](https://github.com/sharkdp/fd) and [`ripgrep`](https://github.com/BurntSushi/ripgrep) use that flag with this meaning.

ericchase:
this flag would be greatly appreciated. i'm not sure when cli devs decided to start searching for .gitignore files by default, but it's such a pain to deal with when there is no off button

PR #918, opened 2024-11-17
Switch to custom glob matching instead of OverrideBuilder

### LINKED ISSUE #869: Glob matching ignores ignore list
If I add a file `test.lua` to my `.styluaignore`, it will correctly get ignored in both of these usages:

```console
$ stylua --respect-ignores test.sua
$ ls *.lua | xargs stylua --respect-ignores
```

Unfortunately it does *not* get ignored if matched in a glob pattern:

````console
$ stylua --respect-ignores -g '*.lua' .
```

I would have expected that scenario to skip the file as well, but it does not. In my case this causes problems in several projects and I have to manually generate lists of files with other tooling and pass them as arguments instead of being able to use glob patterns.

JohnnyMorganz:
Fixes #869


JohnnyMorganz:
Need a solution for negative glob matching.

PR #966, opened 2025-03-27
fix(table,function): doc-commented function as table value

### LINKED ISSUE #942: doc-comment and EmmyLua on function defined as table value
Stylua will format a documented function defined as a table value in a way that EmmyLua complains about `@param` doc-comment.

Before formatting ↓, `---@param n` is correctly recognized by EmmyLua (through lua-language-server). doc-comment lines are indented the same above the `function(…)` line.  No complains from EmmyLua
```
local t = {
	plus_one =
    ---@param n number
	---@return number
    function(n)
		return n + 1
	end,
}
```

After formatting ↓, `---@param n` is moved to the key line, and it's mis-aligned with the rest of the doc-comment lines, in this case the `@return one`. EmmyLua complains `Undefined param n` on `---@param n number` .
```
local t = {
	plus_one = 	---@param n number
	---@return number
function(n)
		return n + 1
	end,
}
```

Expected behavior: doc-comment alignment has higher priority, so all doc-comment lines are kept at same indentation. Don't mind if this results in `function(…)` line having a different indent

FelipeLema:
fixes #942

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ```lua
 local t = {
 	plus_one =
-	---@param n number
-	---@return number
-	function(n)
-		return n + 1
-	end,
+		---@param n number
+		---@return number
+		function(n)
+			return n + 1
+		end,
 }
 ```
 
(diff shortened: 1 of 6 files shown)


FelipeLema:
how do I de-dent the function body? @JohnnyMorganz

FelipeLema:
holy smokes! `format_function_body` is a 217-line code blob

JohnnyMorganz:
> holy smokes! format_function_body is a 217-line code blob

Yeah... some of the code isn't then greatest. Would definitely like to improve some of it. Someday...

I've implemented a fix for this at https://github.com/JohnnyMorganz/StyLua/pull/982

PR #969, opened 2025-04-11
Fix panic when `--respect-ignores --stdin-filepath` on external path to cwd

phanen:
Reproduction:
```sh
echo '/' >.styluaignore
RUST_BACKTRACE=full cargo run -- --respect-ignores --stdin-filepath /tmp/random_path.lua -
```


JohnnyMorganz (approved):
Thank you!

PR #970, opened 2025-04-17
feat: Stylua LSP server

### LINKED ISSUE #936: FR: Make available as LSP
Many formatters/linters such as `biome` or `ruff` are also available as LSP, even though they aren't "real" language servers.

Offering an LSP, even if it is just for the formatting command, has its advantages, as it allows for easier integration into editors. With editors like nvim, for example, integrating `stylua` requires an extra plugin, while an LSP could be used out of the box.

PolyMeilex:
closes #936

For now I did not get around to porting the VsCode extension to LSP, so the easiest way to try this out is to use nvim:
```sh
cargo install --git https://github.com/PolyMeilex/StyLua.git --branch lsp
```
```lua
vim.lsp.config["stylua-lsp"] = {
  cmd = { "stylua", "--lsp" },
  filetypes = { "lua" },
}
vim.lsp.enable("stylua-lsp")
```


PolyMeilex:
Lovely, I have been using this daily for the last 5 months without any issues so I'm happy to see it land 🚀 ❤️

PR #1013, opened 2025-08-12
Move config file resolver to library

flip1995:
When using stylua as a library, the only formatting interface it provides are the `format_code` and `format_ast` functions, which both already expect a config file. To replicate the config file discovery that the stylua CLI implements, one would have to copy this.

My motivation: For one of my projects I'm creating bindings for other languages to the stylua_lib. There it would be really useful, if I could use the config file discovery.


src/cli/config.rs
```diff
@@ -1,347 +0,0 @@
...
-    use clap::StructOpt;
-    use stylua_lib::{CallParenType, IndentType, LineEndings, LuaVersion, QuoteStyle};
-
-    #[test]
-    fn test_override_syntax() {
-        let override_opt = Opt::parse_from(vec!["BINARY_NAME", "--syntax", "Lua51"]);
-        let default_config = Config::new();
-        let config = load_overrides(default_config, &override_opt);
-        assert_eq!(config.syntax, LuaVersion::Lua51);
-    }
```
flip1995, line 298: Those tests are no longer possible, as the `load_overrides` was moved to the lib part. Testing this function there, without involving `Opt::parse_from` seemed pointless to me. So I removed them.

PR #1018, opened 2025-08-29
fix(release-gitter): correct Windows asset name for pip build

izzalDev:
Currently, installing StyLua via pip fails on Windows with the following error:

```
ValueError: Could not find asset named {'stylua-windows-AMD64.zip'} on release v2.1.0
```

This happens because the release-gitter configuration is looking for `stylua-windows-AMD64.zip`, while the actual Windows asset is published under a different name (e.g., `stylua-win64.zip` or `stylua-windows-x86_64.zip`).

This PR updates the release-gitter configuration to match the correct Windows asset name, allowing StyLua to be successfully built and installed via pip on Windows.

---
✅ Fixes pip installation failure on Windows  
✅ Ensures release-gitter resolves the proper asset

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,9 +6,23 @@ build-backend = "pseudo_builder"
 git-url = "https://github.com/JohnnyMorganz/StyLua"
 extract-all = true
 format = "stylua-{system}-{arch}.zip"
+exec = """python -c "
+import os
+import stat
+import platform
+
+f = 'stylua'
+s = platform.system()
+
+if s in ('Linux', 'Darwin') and os.path.exists(f):
+    current_mode = os.stat(f).st_mode
+    os.chmod(f, current_mode | stat.S_IEXEC)
+" """
+
 [tool.release-gitter.map-arch]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -4,11 +4,25 @@ build-backend = "pseudo_builder"
...
+import platform
+
+f = 'stylua'
+s = platform.system()
+
+if s in ('Linux', 'Darwin') and os.path.exists(f):
+    current_mode = os.stat(f).st_mode
+    os.chmod(f, current_mode | stat.S_IEXEC)
+" """
+
```
JohnnyMorganz, line 21: Why do we need all this extraction change?

I'd imagine the only thing needed would be the new entry you've added to `map-arch`

JohnnyMorganz (commented):
Thanks for the contribution!

pyproject.toml
JohnnyMorganz: Makes sense, thanks for clarifying

JohnnyMorganz (approved):

PR #1031, opened 2025-09-15
Smaller text edits in lsp formatting handler

TungstnBallon:
Hi, I saw the `TODO` in the new LSP implementation and thought I'd try it.

Now, the formatting handler does a byte by byte diff of the unformatted and formatted contents, and only responds with the changes between the two.
An alternative would be a line by line diff, but I found that doing it byte-wise has two advantages:
1. `document.position_at()` can be used to compute positions
2. The formatted string can just be sliced to get the new text, no need to search for line endings

<br>

Additionally, I took the liberty of implementing two other nitpicks that I had with the LSP. Let me know if you'd prefer separate PRs or don't want these at all.
1. The servers initialize response includes a [serverInfo](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult) field
3. [FormattingOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions) are respected

Couldn't find a CONTRIBUTING.md, so I hope I didn't miss any PR guidelines 😅.

### DIFF
--- a/src/cli/lsp.rs
+++ b/src/cli/lsp.rs
@@ -4,8 +4,9 @@ use lsp_server::{Connection, ErrorCode, Message, Response};
 use lsp_textdocument::{FullTextDocument, TextDocuments};
 use lsp_types::{
     request::{Formatting, RangeFormatting, Request},
-    DocumentFormattingParams, DocumentRangeFormattingParams, OneOf, Position, Range,
-    ServerCapabilities, TextDocumentSyncCapability, TextDocumentSyncKind, TextEdit, Uri,
+    DocumentFormattingParams, DocumentRangeFormattingParams, InitializeResult, OneOf, Position,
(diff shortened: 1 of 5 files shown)


src/cli/lsp.rs
```diff
@@ -329,17 +435,8 @@ mod tests {
...
-            TextEdit {
-                range: Range {
-                    start: Position::new(0, 0),
-                    end: Position::new(0, 14),
-                },
-                new_text: "local x = 1\n".to_string(),
-            }
-        );
+        let formatted = with_edits(contents, edits);
+        assert_eq!(formatted, "local x = 1\n");
```
JohnnyMorganz, line 439: Just as a sanity check, do you mind also expanding the edits list into assertions of what the edits are composed of? To make it clearer what is the output that the byte diff produces

JohnnyMorganz (approved):
Thank you!

Issue #863, opened 2024-06-03
No uncollapse simple expressions, `collapse_simple_statement` only when already collapsed


d-enk:
I would like simple expressions to can collapse, but complex not

This has been annoying for a very long time
but I was too lazy to write issue)

that is, if there is already a newline no need to collapse

like [gofmt](https://go.dev/play/p/K9a2qmky0TP) behavior

next code can't be changed (additional option may be required)
```lua
if ok then return end

if ok then
  return
end

function func() return end

function func()
  return
end
```

I would also personally like
so that there no new line instead `;` between commands

```lua
some1(); some2(); break
```

but maybe I'm the only one
so for now without a separate issue

toupeira:
I'd also like an option to disable this rule, I found that in my Neovim configuration I often want collapsed functions and conditionals in some contexts, but not in others.

PR #1112, opened 2026-05-15
Add Input mode for collapse_simple_statement

ddashdev:
Adds a new `Input` mode for `collapse_simple_statement`.

It allows Stylua to preserve whether simple functions and conditionals were originally collapsed or expanded, instead of enforcing one global collapse behavior across the whole codebase.

With this mode:
```luau
if x then return end
```
stays collapsed, while:
```luau
if x then
    return
end
```
stays expanded.

The motivation is making it possible for users to keep a mixed style where some simple statements are collapsed and others are intentionally expanded, not all simple statements carry the same importance in a codebase. In some cases a collapsed statement improves readability by keeping a guard clause or empty function compact, while in others, expanding the statement makes the control flow or intent more visible.

The existing `collapse_simple_statement` modes require users to choose one global behavior: collapse matching statements everywhere, or expand them everywhere. This can be limiting when different simple statements benefit from different formatting choices.

And, while `stylua: ignore` comments can work around individual cases, they add noise and become difficult to maintain when used repeatedly for formatting preferences. Adding an `Input` mode follows the existing input-preserving semantics used by other Stylua rules like `call_parenthesis`, and gives users the ability to preserve these choices per situation without changing the default behavior.

**Note:** I used AI assistance while working through parts of the implementation, but reviewed the resulting changes and verified the behavior with tests.


ddashdev:
Also closes #863

PR #1123, opened 2026-05-23
Fix sort_requires not recognizing `const` imports in luau

ddashdev:
## Problem
`sort_requires` only recognized imports parsed as `Stmt::LocalAssignment`, so Luau imports declared with the new `const` keyword were ignored.

For example, this stayed unsorted:

```lua
const Zebra = require("Zebra")
const Apple = require("Apple")
```

## Summary
- Adds Luau-only handling for `Stmt::ConstAssignment` in sort-requires
- Reuses the same require-matching logic for `local` and `const` assignments
- Preserves existing grouping, blank-line splitting, ignored-node behavior, and `game:GetService` sorting
- Adds tests for `const` requires, mixed `local`/`const` blocks, type assertions, and services
- Updates docs and changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


riptscripts:
Plss

